PR #8, opened 2018-02-07
Allow setting the initial commit for the wptsync upstream command.

jgraham:
This is mostly for the transient initial case where we assume that we are up to date with
upstream on central. But we probably actually aren't so in this case we want to allow
running upstreaming over some larger range of commits. Because the base commit isn't
in the normal pulse message we get, we smuggle the base commit though a _wptsync field
when invoking the handler from a command.

### DIFF
--- a/sync/command.py
+++ b/sync/command.py
@@ -72,7 +72,8 @@ def get_parser():
     parser_bug.set_defaults(func=do_bug)
 
     parser_upstream = subparsers.add_parser("upstream", help="Run the upstreaming code")
-    parser_upstream.add_argument("rev", nargs="?", help="Revision to upstream to")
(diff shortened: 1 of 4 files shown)


sync/command.py
```diff
@@ -72,7 +72,8 @@ def get_parser():
     parser_bug.set_defaults(func=do_bug)
 
     parser_upstream = subparsers.add_parser("upstream", help="Run the upstreaming code")
-    parser_upstream.add_argument("rev", nargs="?", help="Revision to upstream to")
+    parser_upstream.add_argument("--base-rev", help="Base revision for upstreaming")
```
mjzffr: You don't want `--base-rev` to be optional?

Issue #16, opened 2018-02-12
Multiple comments about upstream status changing


jgraham:
See e.g. https://bugzilla.mozilla.org/show_bug.cgi?id=1424878#c18

PR #20, opened 2018-02-13
Use combined PR status to determine success. (#16)

mjzffr:
The sequence of GitHub status events that trigger handlers.handle_status
can sometimes include many "success" statuses. We don't want to
report success for each such event, only for the overall success of all
checks. We can do this by requesting combined status from the GitHub
API every time we see a non-"pending" status event.

Similarly, when downstreaming, we should only trigger a try
push once the combined status is "success".

### DIFF
--- a/sync/base.py
+++ b/sync/base.py
@@ -702,6 +702,17 @@ class SyncProcess(object):
         self.data["pr"] = value
 
     @property
+    def last_pr_check(self):
+        return self.data.get("last-pr-check", {})
+
+    @last_pr_check.setter
+    def last_pr_check(self, value):
+        if value is not None:
+            self.data["last-pr-check"] = value
+        else:
+            del self.data["last-pr-check"]
+
+    @property
     def error(self):
         return self.data.get("error")
 
--- a/sync/gh.py
+++ b/sync/gh.py
@@ -35,8 +35,9 @@ class GitHub(object):
         return self._repo
 
     def get_pull(self, id):
+        id = int(id)
         if id not in self.pr_cache:
-            self.pr_cache[id] = self.repo.get_pull(int(id))
+            self.pr_cache[id] = self.repo.get_pull(id)
         return self.pr_cache[id]
 
     def create_pull(self, title, body, base, head):
@@ -80,11 +81,26 @@ class GitHub(object):
         issue = self.repo.get_issue(pr_id)
         issue.add_to_labels(*labels)
 
-    def get_combined_status(self, pr_id):
-        pr = self.get_pull(pr_id)
-        head_commit = self.repo.get_commit(pr.head.ref)
+    @staticmethod
+    def _summary_state(statuses):
+        states = {item.state for item in statuses}
+        overall = None
+        if states == {"success"}:
+            overall = "success"
+        elif "pending" in states or states == set():
+            overall = "pending"
+        if "failure" in states or "error" in states:
+            overall = "failure"
+        return overall
 
-        return head_commit.get_combined_status()
+    def get_combined_status(self, pr_id, exclude=None):
+        if exclude is None:
+            exclude = set()
+        pr = self.get_pull(pr_id)
+        combined = self.repo.get_commit(pr.head.sha).get_combined_status()
+        statuses = [item for item in combined.statuses if item.context not in exclude]
+        state = combined.state if not exclude else GitHub._summary_state(statuses)
+        return state, statuses
 
     def pull_state(self, pr_id):
         pr = self.get_pull(pr_id)
@@ -124,15 +140,6 @@ class GitHub(object):
             pr.url + "/reviews",
             input=post_parameters
         )
-
-    def status_checks_pass(self, pr_id, exclude=None):
-        if exclude is None:
-            exclude = set()
-        statuses = {item.context: item for item in self.get_combined_statuses(pr_id)
-                    if item.context not in exclude}
-        non_success = {context: item for context, item in statuses.iteritems()
-                       if item.state != "success"}
-        return len(non_success) == 0, non_success
 
     def pr_for_commit(self, sha):
         logger.info("Looking up PR for commit %s" % sha)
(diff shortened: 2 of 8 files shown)


mjzffr:
Gah, I forgot that for upstreaming our bot sets a status about whether the patch has landed on gecko.

sync/gh.py
```diff
@@ -126,12 +126,11 @@ def approve_pull(self, pr_id):
...
 
     def status_checks_pass(self, pr_id, exclude=None):
+        c = self.get_combined_status(pr_id)
         if exclude is None:
-            exclude = set()
-        statuses = {item.context: item for item in self.get_combined_statuses(pr_id)
-                    if item.context not in exclude}
-        non_success = {context: item for context, item in statuses.iteritems()
-                       if item.state != "success"}
+            return c.state == "success", [item for item in c.statuses if item.state != "success"]
```
jgraham: I'm not a big fan of special casing the `exclude=None` case like this.

sync/gh.py
```diff
@@ -126,12 +126,11 @@ def approve_pull(self, pr_id):
...
+        statuses = [item for item in c.statuses if item.context not in exclude]
+        non_success = [item for item in statuses if item.state != "success"]
```
jgraham: Isn't this just `non_success = [item for item in statuses if item.context not in exclude and item.state != "success"]`

sync/base.py
```diff
@@ -698,6 +698,17 @@ def pr(self, value):
             raise AttributeError("Can't set attribute")
         self.data["pr"] = value
 
+    @property
+    def last_pr_check(self, value):
```
jgraham: I think this should be called `last_pr_state`

sync/upstream.py
```diff
@@ -371,11 +371,11 @@ def try_land_pr(self):
 
         msg = None
         all_status_success, non_success = env.gh_wpt.status_checks_pass(
-            self.pr, exclude="upstream/gecko")
+            self.pr, exclude=["upstream/gecko"])
```
jgraham: Make that a set

sync/upstream.py
```diff
@@ -721,13 +721,22 @@ def status_changed(git_gecko, git_wpt, sync, context, status, url, sha):
     if status == "pending":
         # Never change anything for pending
         return landed
-
-    if status == "success":
+    previous_status = sync.last_pr_check
+    success, non_success = env.gh_wpt.status_checks_pass(
+        sync.pr,
+        exclude=["upstream/gecko"]
```
jgraham: set

sync/upstream.py
```diff
@@ -721,13 +721,22 @@ def status_changed(git_gecko, git_wpt, sync, context, status, url, sha):
     if status == "pending":
         # Never change anything for pending
         return landed
-
-    if status == "success":
+    previous_status = sync.last_pr_check
+    success, non_success = env.gh_wpt.status_checks_pass(
```
jgraham: I think this whole bit would be easier if it used `get_combined_status` directly (maybe allow `exclude` to be passed into that one too).

sync/upstream.py
```diff
@@ -721,13 +721,22 @@ def status_changed(git_gecko, git_wpt, sync, context, status, url, sha):
...
+        sync.pr,
+        exclude=["upstream/gecko"]
+    )
+    if success:
+        sync.last_pr_check = "success"
+    else:
+        states = {item.state for item in non_success}
+        sync.last_pr_check = "failure" if "failure" in states else "pending"
+    if success and previous_status != "success":
```
jgraham: I think we should try the landing irrespective of the previous status, but only  add a comment if something changed.

sync/upstream.py
```diff
@@ -721,13 +721,22 @@ def status_changed(git_gecko, git_wpt, sync, context, status, url, sha):
...
         sync.error = None
         if sync.gecko_landed():
             landed = sync.try_land_pr()
         else:
-            env.bz.comment(sync.bug, "Upstream web-platform-tests status checked passed, "
+            env.bz.comment(sync.bug, "Upstream web-platform-tests status checks passed, "
```
jgraham: Ideally we only add this comment if we didn't already. Maybe not worth worrying about in this PR.

sync/base.py
jgraham: (also shouldn't have a `value` parameter ofc)

jgraham:
It occured to me in the middle of the night that we should store the status and the commit sha, because if the sha changes we want to rerun the tasks even if the status didn't change.

sync/upstream.py
mjzffr: I think with this new approach, repeated comments can still happen but will be much less likely than before in practice.

sync/upstream.py
```diff
@@ -370,19 +370,19 @@ def try_land_pr(self):
...
-                details = "\n".join(details)
-                msg = ("Can't merge web-platform-tests PR due to failing upstream checks:\n%s" %
-                       details)
+        check_state, statuses = env.gh_wpt.get_combined_status(
+            self.pr, exclude={"upstream/gecko"})
+        non_success = [i for i in statuses if i.state != "success"]
+        if check_state not in ["success", "pending"]:
+            details = ["Github PR %s" % env.gh_wpt.pr_url(self.pr)]
+            for item in non_success:
+                if item.state == "pending":
```
jgraham: Is this actually possible now? Should it be "failing" instead?

sync/upstream.py
```diff
@@ -721,13 +721,16 @@ def status_changed(git_gecko, git_wpt, sync, context, status, url, sha):
...
         if sync.gecko_landed():
             landed = sync.try_land_pr()
-        else:
-            env.bz.comment(sync.bug, "Upstream web-platform-tests status checked passed, "
+        elif previous_check != sync.last_pr_check:
+            env.bz.comment(sync.bug,
+                           "Upstream web-platform-tests status checks passed, "
                            "PR will merge once commit reaches central.")
     else:
         env.bz.comment(sync.bug, "Upstream web-platform-tests status %s for %s. "
```
jgraham: This comment should only happen if the `previous_check != sync.last_pr_check` condition holds, correct? Should we just add the message here about which tests fail rather than in `try_land_pr` above (and not add anything to the bug in that case)?

sync/upstream.py
mjzffr: If the overall state is "failure" (check_state), then it's still possible to have "pending" in the list of individual statuses. (>=1 failure/error implies overall failure).

So what I'm going for is to report the failures/errors in the message details.

PR #22, opened 2018-02-14
Add facility to wait for a specific gecko hg rev to appear when updating repos.

jgraham:
I think this is required because there's a synchronisation delay between mozilla-unified and
the source repos. We want to deal with unified, but the events we are getting are based on the
source repos, so the commits haven't necessarily appeared when we run the handlers.

### DIFF
--- a/sync/gitutils.py
+++ b/sync/gitutils.py
@@ -1,3 +1,5 @@
+import time
+
 import git
 import log
 from env import Environment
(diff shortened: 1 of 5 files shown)


sync/gitutils.py
```diff
@@ -8,21 +10,53 @@
...
-        git_wpt.git.fetch("origin")
+        _update_wpt(git_wpt)
+
+
+def until(func, cond, max_tries=5):
+    for i in xrange(max_tries):
+        func()
+        if cond():
+            break
+        time.sleep(1 * (i + 1))
```
mjzffr: Why `1 *`?

PR #23, opened 2018-02-14
Use combined upstream status checks when checking if we can merge a PR.

jgraham:
This avoids the problem where we misuse the status API and think that
something has failed when it hasn't. We also add better logging for
the failing case.

### DIFF
--- a/sync/gh.py
+++ b/sync/gh.py
@@ -80,11 +80,11 @@ class GitHub(object):
         issue = self.repo.get_issue(pr_id)
         issue.add_to_labels(*labels)
 
-    def get_statuses(self, pr_id):
+    def get_combined_status(self, pr_id):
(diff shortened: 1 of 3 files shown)


sync/gh.py
```diff
@@ -125,17 +125,14 @@ def approve_pull(self, pr_id):
...
-            return False
-        statuses = self.get_statuses(pr_id)
-        latest = {}
-        for status in statuses:
-            if status.context not in latest:
-                latest[status.context] = status.state
-
-        return all(status == "success" for status in latest.itervalues())
+    def status_checks_pass(self, pr_id, exclude=None):
+        if exclude is None:
```
mjzffr: The only major difference with my version in #20 is that I call a different API endpoint to get summary status directly from GitHub if we're not excluding any contexts.

PR #24, opened 2018-02-14
Document what is needed from fresh checkout

AutomatedTester:


bin/run_docker_dev.sh
```diff
@@ -1,16 +1,18 @@
 #!/bin/bash
 
-# Helper script for running docker in a dev environment. 
+# Helper script for running docker in a dev environment.
 # Use run_docker.sh instead in prod
 
 if [[ $1 == "build" ]]; then
+    mkdir -p devenv/
+    cp test/testdata/* devenv/
```
mjzffr: I'd only make the copy if `devenv` is empty. Once you've customized your config, you won't want to always overwrite it with test data.

mjzffr:
@AutomatedTester I'll wait to merge until you tell me that you got run_docker_env.sh --shell working, since you might want to add a couple more notes as a result.

bin/run_docker_dev.sh
```diff
@@ -1,16 +1,37 @@
 #!/bin/bash
 
-# Helper script for running docker in a dev environment. 
+# Helper script for running docker in a dev environment.
 # Use run_docker.sh instead in prod
 
 if [[ $1 == "build" ]]; then
+    {
+      echo "Creating workspace\n"
```
mjzffr: Creating devenv\n

bin/run_docker_dev.sh
```diff
@@ -1,16 +1,37 @@
...
+      mkdir devenv/
+      echo "Copying configs to workspace"
```
mjzffr: to devenv

PR #30, opened 2018-02-15
Set bug priority as P3 on creation.

jgraham:
Also add a note that the bug will be auto-merged. This is for better
compatibility with Gecko triage processes.


jgraham:
Fixes #31

PR #62, opened 2018-02-22
Fix up the handling of non-synced files.

jgraham:
The previous approach to non-synced files didn't actually work because
the patterns were only matched against the filename and not the full
path. Also the files were deleted because we didn't restore the copy
from the current gecko tree.

In this new approach we match the full path for all the files with a
custom ignore function, and use git checkout_index on the gecko side
to ensure that the files are later restored.

### DIFF
--- a/sync/landing.py
+++ b/sync/landing.py
@@ -1,6 +1,7 @@
 import re
 import os
 import shutil
+from collections import defaultdict
 
 import git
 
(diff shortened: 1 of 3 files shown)


sync/landing.py
```diff
@@ -130,17 +132,35 @@ def add_pr(self, pr_id, wpt_commits):
...
-        # Some files we don't want to update on import ever
-        ignore_patterns = ["LICENSE", ".git", "resources/testdriver-vendor.js"]
+        ignore_paths = defaultdict(set)
+        for name in keep_files | ignore_files:
+            src, name = os.path.split(os.path.join(src_path, name))
+            ignore_paths[src].add(name)
 
-        shutil.copytree(git_work_wpt.working_dir,
+        def ignore_names(src, names):
+            logger.info("Copying %s %s" % (src, names))
```
mjzffr: That's going to be a lot of logging -- once per subdir.

PR #63, opened 2018-02-22
Clean up commits created during landing.

jgraham:
Ensure we create only a single commit inclusingn any additional
metadatta updates and the change in the in-tree sync point.

### DIFF
--- a/sync/landing.py
+++ b/sync/landing.py
@@ -370,26 +370,26 @@ Automatic update from web-platform-tests%s
                 self.add_metadata(sync)
 
     @property
-    def metadata_commit(self):
-        if self.gecko_commits[-1].metadata["wpt-type"] == "metadata":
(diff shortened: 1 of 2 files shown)


sync/landing.py
```diff
@@ -347,26 +347,26 @@ def apply_prs(self, landable_commits):
                 self.add_metadata(sync)
 
     @property
-    def metadata_commit(self):
-        if self.gecko_commits[-1].metadata["wpt-type"] == "metadata":
+    def landing_commit(self):
+        head = self.gecko_commits.head
+        if (head.metadata.get("wpt-type") == "landing" and
+            head.metadata.get("wpt-head") == self.wpt_commits.head.sha1):
             return self.gecko_commits[-1]
```
mjzffr: nit: I think it would be clearer to return `head`

PR #71, opened 2018-02-27
Don't try to upstream commits resulting from a landing.

jgraham:
The upstreaming code was accidentially trying to re-upstream stuff
that itself came from upstream. This patch checks if the commit has
metadata indicating that it came from upstream, and if it does we
filter it out of the commits considered for upstreaming.

### DIFF
--- a/sync/upstream.py
+++ b/sync/upstream.py
@@ -10,6 +10,7 @@ from mozautomation import commitparser
 import base
 import log
 import commit as sync_commit
+from downstream import DownstreamSync
 from gitutils import update_repositories, gecko_repo
 from pipeline import AbortError
 from env import Environment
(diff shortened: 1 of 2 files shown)


sync/upstream.py
```diff
@@ -443,7 +448,9 @@ def wpt_commits(git_gecko, first_commit, head_commit):
                                       paths=env.config["gecko"]["path"]["wpt"],
                                       reverse=True,
                                       max_parents=1)]
-    return [item for item in commits if not item.metadata.get("wptsync-skip")]
+    return [commit for commit in commits
+            if not commit.metadata.get("wptsync-skip") or
```
mjzffr: @jgraham Where does `wptsync-skip` get set? I can't find any other occurrences.

jgraham: You can set it in a commit message if the commit shouldn't be udpatreamed for whatever reason.

PR #76, opened 2018-03-01
Fixup typo in decision task error message

jgraham:


jgraham:
This is trivial, so just landing it.

PR #82, opened 2018-03-02
Handle empty metadata commits.

jgraham:
If we are trying to cherry-pick metadata and get an empty commit, we
currently error out. There are two possible ways to handle this;
either pass --allow-empty and so end up with a meaningless commit, or
handle the error and continue without making a commit. The latter
seems preferable in terms of not having useless commits in the
history, so we do that here. Unfortunately it's a little fragile since
we depend on the error message string to determine what happened.

### DIFF
--- a/sync/landing.py
+++ b/sync/landing.py
@@ -319,8 +319,24 @@ Automatic update from web-platform-tests%s
     def add_metadata(self, sync):
         if self.has_metadata(sync):
             return
-        if sync.metadata_commit:
-            self.gecko_worktree.get().git.cherry_pick(sync.metadata_commit.sha1)
+        if sync.metadata_commit and not sync.metadata_commit.is_empty():
+            worktree = self.gecko_worktree.get()
+            try:
+                worktree.git.cherry_pick(sync.metadata_commit.sha1)
+            except git.GitCommandError as e:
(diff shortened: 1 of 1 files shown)


sync/landing.py
```diff
@@ -319,8 +319,18 @@ def has_metadata(self, sync):
...
         if self.has_metadata(sync):
             return
-        if sync.metadata_commit:
-            self.gecko_worktree.get().git.cherry_pick(sync.metadata_commit.sha1)
+        if sync.metadata_commit and not sync.metadata_commit.is_empty():
+            worktree = self.gecko_worktree.get()
+            try:
+                worktree.git.cherry_pick(sync.metadata_commit.sha1)
+            except git.GitCommandError as e:
+                if (str(e.status) == "1" and
```
AutomatedTester: what does `"1"` mean? Can we avoid a magic number here?

jgraham: It's the status code that git exits with in this case. I don't think there's a reasonable way to avoid a magic number; definine `GIT_CHERRY_PICK_EXIT_CODE_ON_EMPTY_COMMIT` doesn't seem helpful. But I could add a comment.

AutomatedTester: Let's add a comment just so that we know what it means in the future

PR #93, opened 2018-03-06
Update wpt commits correctly after a merge

jgraham:
After wpt commits are either merged or fast-forwarded onto master the
range origin/master..origin/pr/<id> is empty since the endpoint is an
ancestor of the master branch. Instead we should take the commits that
actually landed onto the master branch explicitly, setting the base
commit to whatever the commit prior to landing was.

The way we handle updating the base commit is non-ideal since it has
to be set in the sync data and then updated in the CommitRange
object. However this should be enough to fix the problem we currently have.

### DIFF
--- a/sync/downstream.py
+++ b/sync/downstream.py
@@ -141,9 +141,22 @@ class DownstreamSync(base.SyncProcess):
             self.status = "open"
             self.pr_status = "open"
 
-    def update_wpt_head(self):
+    def update_wpt_commits(self):
         if not self.wpt_commits.head or self.wpt_commits.head.sha1 != self.pr_head.sha1:
             self.wpt_commits.head = self.pr_head
+        if len(self.wpt_commits) == 0 and self.git_wpt.is_ancestor(self.wpt_commits.head.sha1,
+                                                                   "origin/master"):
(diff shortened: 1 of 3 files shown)


sync/downstream.py
```diff
@@ -141,9 +141,22 @@ def update_status(self, action, merge_sha=None, wpt_base=None):
...
-    def update_wpt_head(self):
+    def update_wpt_commits(self):
         if not self.wpt_commits.head or self.wpt_commits.head.sha1 != self.pr_head.sha1:
             self.wpt_commits.head = self.pr_head
+        if len(self.wpt_commits) == 0 and self.git_wpt.is_ancestor(self.wpt_commits.head.sha1,
+                                                                   "origin/master"):
+            # The commits landed on master so we need to change the commit
+            # range to not use origin/master as a base
+            base_commit = None
+            assert self.wpt_commits.head.pr() == self.pr
```
AutomatedTester: Do we need some docs on what to do if this assertion is hit?

jgraham: I don't really know what they would say. That would indicate a bug because we would be getting the wrong commits for a pr.

PR #103, opened 2018-03-09
Add clean-up to entrypoint script

mjzffr:
With these changes, `docker stop -t 60 <container>` will give celery and rabbitmq a chance to shutdown gracefully. The entrypoint script also attempts to clean up stale pid files when starting up.

### DIFF
--- a/docker/Dockerfile.dev
+++ b/docker/Dockerfile.dev
@@ -17,6 +17,11 @@ RUN set -eux; \
     apt-get update; \
     apt-get install -y git sudo less emacs-nox wget gnupg apt-transport-https
 
+# install tini for process management so we can use its -g option
+ENV TINI_VERSION v0.17.0
+ADD https://github.com/krallin/tini/releases/download/${TINI_VERSION}/tini /tini
+RUN chmod +x /tini
+
 ### Install latest erlang
 # add buster sources.list
 RUN set -eux; \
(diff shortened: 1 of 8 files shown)


docker/Dockerfile.dev
```diff
@@ -5,6 +5,12 @@ RUN groupadd --gid 10001 app && \
 
 ENV RABBITMQ_CONFIG_FILE=/etc/rabbitmq/rabbitmq
 COPY ./docker/rabbitmq.conf /etc/rabbitmq/rabbitmq.conf
+RUN echo LOG_BASE=/app/workspace/logs > /etc/rabbitmq/rabbitmq-env.conf
+
+# install tini for process management
+ENV TINI_VERSION v0.17.0
```
jgraham: I don't entirely understand why we can't use the version that ships  with Docker, but I asume there's a good reason. Might be nice to add a comment for this.

mjzffr: @jgraham Will do. 

The version that ships with docker is only used via the `--init` option of  `docker run`, which is what we were doing before. If we want to customize how tini is called (in our case with `-g`) there's no way to do that with the built-in version, so a common thing to do is to include it in your image.

(Another reason people include it in their images is that the built-in option isn't available with things like docker-compose and other container orchestration software.)

PR #107, opened 2018-03-12
Add an option to include incomplete syncs when landing.

jgraham:
This is helpful in catching up with a backlog of PRs that could build
up for various reasons. The metadata update will all happen in the
landing, so it does have the disadvantage that we won't detect or
disable unstable tests or get per-PR metadata, so it's a method of
last resort, but it can nevertheless be useful.


sync/command.py
```diff
@@ -126,6 +128,8 @@ def get_parser():
     parser_landable.add_argument("--retrigger", action="store_true", default=False,
                                  help="Try to update all unlanded PRs that aren't Ready "
                                  "(requires --all)")
+    parser_landable.add_argument("--include-incomplete", action="store_true", default=False,
```
AutomatedTester: why do we have this twice?

Issue #112, opened 2018-03-14
Wpt-sync should preserve authorship information.


emilio:
Probably the author / authors of the patch should be the authors of the commit.

emilio:
Fixed by #115, thanks again James!

PR #115, opened 2018-03-15
By default preserve author when moving commits.

jgraham:
This should ensure that the author is preserved when upstreaming.


jgraham:
Fixes #112

PR #126, opened 2018-03-26
Exclude intermittent failures from landing metadata update 

mjzffr:
wpt failures found in final try push for landing process are
retriggered to determine whether they are intermittent. Any
retriggered test that mostly passes is excluded from the
metadata update.

Also set the landing status to "error" if too many wpt tasks
are failing.

@jgraham I need to add some unit tests and do some more testing locally, but this is pretty much ready for review.


sync/trypush.py
```diff
@@ -389,29 +392,67 @@ def wpt_tasks(self, force_update=False):
         self._data["tasks"] = wpt_tasks
         return wpt_tasks
 
+    def retrigger_failures(self, count=_retrigger_count):
+        wpt_tasks = self.wpt_tasks()
+        failures = [task.get("status", {}).get("state", None) != "success" for task in wpt_tasks]
```
jgraham: I feat that will try to retrigger things that are unscheduled. Isn't "failed" the only state we want to retrigger?

sync/trypush.py
```diff
@@ -389,29 +392,67 @@ def wpt_tasks(self, force_update=False):
...
     def success(self):
         """Check if all the wpt tasks in a try push ended with a successful status"""
         wpt_tasks = self.wpt_tasks()
         return all(task.get("status", {}).get("state", None) == "success" for task in wpt_tasks)
 
-    def download_logs(self, raw=True, report=True):
+    def success_rate(self):
+        wpt_tasks = self.wpt_tasks()
+        success = [task.get("status", {}).get("state", None) == "success" for task in wpt_tasks]
+        return len(success) * 1.0 / len(wpt_tasks)
```
jgraham: Let's spell this `float(len(success)) / len(wpt_tasks)`

sync/landing.py
```diff
@@ -670,7 +670,32 @@ def land_to_gecko(git_gecko, git_wpt, prev_wpt_head=None, new_wpt_head=None,
...
+            return
+        num_new_jobs = try_push.retrigger_failures()
+        logger.info("%s new tasks scheduled on try for %s" (num_new_jobs, sync.bug))
+        if num_new_jobs:
+            env.bz.comment(sync.bug,
+                           ("Retriggered failing web-platform-test tasks on "
+                            "try before final metadata update."))
+            return
+    for name, data in retriggered.iteritems():
+        total = sum(data["states"].itervalues()) * 1.0
```
jgraham: So this doesn't quite work. The problem is that if we have a test that reliably fails on OSX (for example) but passes on Linux then afaict we will not download the logs for the Linux run, and so when we do the metadata update it will conclude that the test fails on all platforms.

I think we need to arrange things so that we have at least one run of each test run on each platform in the logs we download.

A second problem is what about cases where we get an intermittent mixed in with some real failures. We possibly need to update the metadata update code so that for tests with multiple results we only consider the most common result.

sync/trypush.py
```diff
@@ -7,7 +7,7 @@
 
 import base
 import log
-import taskcluster
+import tc_request as tc
```
jgraham: It seems odd to alias this just in one place.

sync/trypush.py
mjzffr: I guess failed or exception. You're right that this might pick up stuff like unscheduled and pending.

sync/landing.py
mjzffr: I'll change the exclusion to keep any logs that represent a "success".

As for the mix of intermittent and real failures, I'll follow up upstream.

sync/landing.py
```diff
@@ -670,7 +670,32 @@ def land_to_gecko(git_gecko, git_wpt, prev_wpt_head=None, new_wpt_head=None,
...
 @base.entry_point("landing")
 def try_push_complete(git_gecko, git_wpt, try_push, sync, allow_push=True):
-    log_files = try_push.download_raw_logs()
+    # min rate of job success to proceed with metadata update
+    target_rate = 0.7
+    retriggered = try_push.retriggered_wpt_states(force_update=True)
+    intermittents = []
+    if not try_push.success and not retriggered:
+        if try_push.success_rate < target_rate:
+            sync.error = (
```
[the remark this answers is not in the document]
mjzffr: Oh, I was under the impression that assigning to `sync.error` results in a bug comment.

mjzffr:
Oh I think I'm going to have to change "success" to "completed" when referring to task state. There's no such thing as "success" in TC. I'll verify tomorrow.

requirements/dev.txt
```diff
@@ -27,7 +27,7 @@ pytest-cov==2.5.1
 pytest-env==0.6.2
 scandir==1.6
 simplegeneric==0.8.1
-six==1.11.0
+six==1.10.0
```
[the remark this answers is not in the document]
jgraham: Oh how silly. OK.

sync/handlers.py
```diff
@@ -3,14 +3,13 @@
 import downstream
 import log
 import landing
-import taskcluster
+import tc_request as tc
```
jgraham: If we're doing this everywhere, let's just call the file tc?

sync/trypush.py
```diff
@@ -231,6 +233,7 @@ class TryPush(base.ProcessData):
     """
     obj_type = "try"
     statuses = ("open", "complete", "infra-fail")
+    _retrigger_count = 9
```
jgraham: Maybe try 4 as a default.

mjzffr: Sure

sync/tc.py
```diff
@@ -15,11 +17,65 @@
...
+RUNNING = "running"
+PENDING = "pending"
+
 logger = log.get_logger(__name__)
 
 env = Environment()
 
 
+class TaskclusterClient(object):
+
```
jgraham: Nit: I don't think we usually have a blank line here, unless there's a docstring or some class attributes above it.

PR #128, opened 2018-03-28
Use wpt-merged status for upstream syncs that have not yet relanded

jgraham:


sync/base.py
```diff
@@ -557,6 +557,8 @@ class SyncProcess(object):
     obj_type = "sync"
     sync_type = "*"
     obj_id = None  # Either "bug" or "pr"
+    statuses = ()
+    status_transitions = []
```
mjzffr: @jgraham I wonder, does anything prevent us from setting a sync as "complete" while self.error is truthy? I don't see any code paths that depend on this, but it shouldn't made sense in any case.

jgraham: Probably not, but let's open another issue for that.

PR #129, opened 2018-03-28
Fix metadata update handing.

jgraham:
Before we would always append metadata to commit messages, resulting
in duplicate metadata in some cases. Instead we want to read the
existing metadata and update it. To make this work we solidify the
definition of metadata:

* It is a line of the form 'foo: bar' i.e. containing a colon followed
  by a space.
* Metadata lines appear at the end of a commit message possibly
  interspersed with blank lines.

This means that any line which happens to look like metadata, but
which appears above a line that does not look like metadata and is not
blank will not be considered metadata.

### DIFF
--- a/sync/commit.py
+++ b/sync/commit.py
@@ -26,11 +26,13 @@ class ApplyError(Exception):
 def get_metadata(text):
     data = {}
     for line in reversed(text.splitlines()):
-        if line:
+        if line.strip():
             m = METADATA_RE.match(line)
             if m:
                 key, value = m.groups()
                 data[key] = value
+            else:
+                break
     return data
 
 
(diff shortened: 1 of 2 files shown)


sync/commit.py
```diff
@@ -134,10 +136,38 @@ def create(cls, repo, msg, metadata, author=None, amend=False):
...
+
+        msg_lines = msg_lines[:max(keep_lines, 1)]
+
         if metadata:
-            metadata_str = "\n".join("%s: %s" % item for item in sorted(metadata.items()))
-            msg = "%s\n%s" % (msg, metadata_str)
-        return msg
+            msg_lines.append("")
+
+        metadata_added = set()
```
mjzffr: I think this can be simplified. Judging by the way `test_metadata_update` is written, we always replace the commit message metadata entirely with what's in `metadata`, so it would work to just write everything in `metadata` right after the "kept lines"? That is, I don't think we actually need to keep track of `metadata_seen` unless it's important to maintain the same order as before...

PR #132, opened 2018-03-28
Allow updating a subset of syncs from GitHub.

jgraham:
This means making multiple API calls, which might hit rate limits. But there doesn't
seem to be an easy way to get all the data for a set of PR ids

### DIFF
--- a/sync/command.py
+++ b/sync/command.py
@@ -32,6 +32,9 @@ def get_parser():
 
     parser_update = subparsers.add_parser("update",
                                           help="Update the local state by reading from GH + etc.")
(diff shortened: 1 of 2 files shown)


sync/command.py
```diff
@@ -32,6 +32,9 @@ def get_parser():
 
     parser_update = subparsers.add_parser("update",
                                           help="Update the local state by reading from GH + etc.")
+    parser_update.add_argument("--sync-type", nargs="*", help="Type of sync to update",
+                               choices=["upstream", "downstream"])
+    parser_update.add_argument("--status", nargs="*", help="Statuses of syncs to update")
```
mjzffr: Could you add an example to the `help` -- "e.g. open"?

Issue #116, opened 2018-03-16
PR titles and commits frequently have trailing punctuation


gsnedders:
See e.g. https://github.com/w3c/web-platform-tests/pull/10074 or https://github.com/w3c/web-platform-tests/pull/9497 or https://github.com/w3c/web-platform-tests/pull/9571.

All of these had r=x in the commit summary, which wpt-sync stripped, but left the trailing punctuation delimiting it from the actual summary.

PR #139, opened 2018-04-04
Strip trailing punctuation in upstreamed PR title (#116)

mjzffr:


sync/upstream.py
```diff
@@ -447,6 +447,7 @@ def commit_message_filter(msg):
         metadata["gecko-reviewers"] = reviewers
     msg = commitparser.replace_reviewers(msg, "")
     msg = commitparser.strip_commit_metadata(msg)
+    msg = msg.rstrip(":;,.")
```
jgraham: I think this should work on the first line of the message, rather than on the whole thing. Also rather than hardcoding a few characters we can use a regxp like `msg.replace("", re.compile("\W*$"), re.UNICODE)` (totally untested and so wrong). But perhaps there's a good reason not to do that e.g. do we want to strip trailing full stops or question marks? Maybe it should just be some combining characters like `:` or `-` or `_`.

sync/upstream.py
```diff
@@ -339,6 +339,7 @@ def create_pr(self):
             time.sleep(1)
 
         commit_summary = self.wpt_commits[0].commit.summary
+        commit_summary = commit_summary.rstrip("!#$%&(*+,-/:;<=>@[\\^_`{|~")
```
mjzffr: I just used whatever is in `string.punctuation` minus quotes, closing parens, period, ?

jgraham: It's probably not unicode correct, but it's also probably good enough :)

But this only updates the GH summary, not the actual commit message, right?

mjzffr: Oh, oops, I thought only the PR title was upsetting. Will take another look.

PR #141, opened 2018-04-05
Retry Pulse consumer connection on socket.error

mjzffr:
Whenever the wptsync service falls over, we get a broken-pipe
socket error originating in the listen_forever code.
At the same time, rabbitmq reports that the client closed
the TCP connection, so what we've been calling "rabbitmq falling over"
actually seems to be our pulse listener falling over. Retrying
the consumer connection is my naive attempt to make this go away.

### DIFF
--- a/sync/listen.py
+++ b/sync/listen.py
@@ -1,6 +1,8 @@
 import os
 import socket
 import urlparse
+import traceback
+from time import sleep
 
 import kombu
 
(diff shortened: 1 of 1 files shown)


sync/listen.py
```diff
@@ -39,26 +41,24 @@ def add_callback(self, exchange, func):
...
-
-    def listen_forever(self):
+    def listen_forever(self, retries=5):
         """Listen for and handle messages until interrupted."""
         if not self._entered:
             raise Exception('must enter context manager before calling')
-
-        while True:
+        pause = 1
+        while True and retries > 0:
```
jgraham: Should we reset retries on a successful connection?

mjzffr: That sounds like a good idea.

sync/listen.py
```diff
@@ -39,26 +41,26 @@ def add_callback(self, exchange, func):
...
+        count = retries
+        while True and count > 0:
             try:
                 self._conn.drain_events(timeout=1.0)
-            except socket.timeout:
-                pass
+                pause = 1
+                count = retries
+            except socket.timeout as e:
+                logger.warning(traceback.format_exc(e))
```
jgraham: Is a warning here sensible? It seems like we might reasonably have no events for 1s?

PR #143, opened 2018-04-06
Mark syncs for PRs that are reverted, and the reverting sync as skip

jgraham:
PRs that are later reverted cause unnecessary work since the revert
won't apply until mozilla/central is updated with the original commit,
which is presumbably broken in some way. Instead of figuring all of
this out, it makes more sense to just skip both the original PR and
the reverting PR, assuming that they overall have no effect.

Putting this code at the start of  update_commits is not ideal, but it
must certainly run before that method since it will typically fail if
the revert doesn't apply to central.

### DIFF
--- a/sync/downstream.py
+++ b/sync/downstream.py
@@ -7,8 +7,11 @@ from __future__ import print_function
 """Functionality to support VCS syncing for WPT."""
 
 import os
+import re
 import traceback
 from collections import defaultdict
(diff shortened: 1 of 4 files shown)


jgraham:
Yes, but I think I'd rather just assume it's right and catch any errors in the landing code, since it's a little simpler.

PR #155, opened 2018-04-12
Account for infra failures when creating or using try pushes.

mjzffr:
This is a tentative PR. I don't have a full picture of the impact on landing code, so would like some initial feedback on whether the overall approach makes sense. 

The rough idea is to mark a sync with "error" if we try to get metadata from an empty try push (no wpt tasks) -- this can happen if tasks expire from taskcluster, which they do after 28 days, or if the decision task failed.

The other piece is to create a try push again during downstreaming if we know the previous one was an infra failure. 

Also note to self before I run out the door: todo - check wpt_tasks in next_try_push, force True.

### DIFF
--- a/sync/downstream.py
+++ b/sync/downstream.py
@@ -107,7 +107,9 @@ class DownstreamSync(base.SyncProcess):
         latest_try_push = self.latest_valid_try_push
         if not latest_try_push:
             return False
-        if latest_try_push.status != "complete":
+        if latest_try_push.status != "complete" or latest_try_push.infra_fail:
+            return False
+        if latest_try_push.expired():
             return False
         return latest_try_push.stability or not self.requires_stability_try
 
@@ -148,6 +150,8 @@ class DownstreamSync(base.SyncProcess):
         so we always assume that any try push is valid"""
 
         latest_try_push = self.latest_try_push
+        if latest_try_push and latest_try_push.expired():
+            latest_try_push = None
 
         # Check if the try push is for the current PR head
         if (latest_try_push and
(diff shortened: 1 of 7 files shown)


sync/downstream.py
```diff
@@ -184,12 +184,22 @@ def next_try_push(self, try_cls=trypush.TrySyntaxCommit):
 
         latest_try_push = self.latest_valid_try_push
 
-        if latest_try_push and latest_try_push.status != "complete":
-            return
+        if latest_try_push:
+            if latest_try_push.status != "complete":
+                return
+            if len(self.try_pushes()) > 20 and latest_try_push.infra_fail:
```
jgraham: 20 does seem like a lot, but if a branch has a lot of changes this could be hit with only one infra fail. We probably want some more nuanced check here.

sync/downstream.py
```diff
@@ -184,12 +184,22 @@ def next_try_push(self, try_cls=trypush.TrySyntaxCommit):
 
         latest_try_push = self.latest_valid_try_push
```
jgraham: Note to self: return if the status isn't open.

sync/downstream.py
```diff
@@ -597,6 +607,8 @@ def try_push_complete(git_gecko, git_wpt, try_push, sync):
         disabled = []
         if not try_push.success():
             if sync.affected_tests:
+                if try_push.failure_limit_exceeded(sync):
```
jgraham: This seems wrong; it's perfectly fine for a non-stability push to fail all jobs.

sync/downstream.py
mjzffr: So my only idea is to count the number of consecutive try pushes that have been infra_fails. That should be conservative enough, yeah?

sync/downstream.py
mjzffr: Instead I added a check for exceptions only to `try_push.wpt_tasks`. Less wrong?

mjzffr:
I'll need to write some new tests, too, of course.

sync/downstream.py
```diff
@@ -107,7 +107,7 @@ def metadata_ready(self):
         latest_try_push = self.latest_valid_try_push
         if not latest_try_push:
             return False
-        if latest_try_push.status != "complete":
+        if latest_try_push.status != "complete" or latest_try_push.infra_fail:
```
jgraham: Maybe we also want to check for expired here?

sync/handlers.py
```diff
@@ -193,13 +193,14 @@ def __call__(self, git_gecko, git_wpt, body):
         logger.info("Setting taskgroup id for try push %r to %s" % (try_push, task_id))
         try_push.taskgroup_id = task_id
 
-        if result != "success":
+        if result != "success" or len(try_push.failed_builds):
             try_push.status = "complete"
             try_push.infra_fail = True
             sync = try_push.sync(git_gecko, git_wpt)
             logger.error("Decision task got status %s for task %s%s" %
```
jgraham: This logging isn't true, right? It's not the status of the decision task.

mjzffr: TaskHandler runs when we get a taskgroup-resolved message about the decision task, so `body["result"]` is just the result of the decision task. But I will add an extra component to the message about build failures.

sync/downstream.py
mjzffr: Done.

PR #158, opened 2018-04-14
Refactor Pulse listener to use kombu's ConsumerMixin

mjzffr:
ConsumerMixin takes care of reviving dropped connections.

See https://github.com/celery/kombu/blob/v4.1.0/kombu/mixins.py#L35

### DIFF
--- a/sync/listen.py
+++ b/sync/listen.py
@@ -1,10 +1,9 @@
 import os
-import socket
 import urlparse
-import traceback
-from time import sleep
 
 import kombu
+
+from kombu.mixins import ConsumerMixin
 
 import log
 import handlers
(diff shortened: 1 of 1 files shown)


mjzffr:
I ran this in prod for about an hour and it seemed alright for about 20 minutes then stopped processing messages and looked stuck in a heartbeats. No pulse-related bugs were logged. I'll look into config options for the mixin.

mjzffr:
I moved the call to queue.queue_declare() and changed how the kombu Consumer is instantiated. Tested locally and it was consuming messages galore! :)

sync/listen.py
```diff
@@ -16,51 +15,26 @@
 logger = log.get_logger(__name__)
 
 
-class Consumer(object):
-    """Represents a Pulse consumer to version control data."""
-    def __init__(self, conn, exchanges, extra_data):
-        self._conn = conn
-        self._consumer = None
-        self._entered = False
+class Consumer(ConsumerMixin):
```
jgraham: So, calling this Consumer, but having it be different from the `Consumer` we create in `get_consumers` is a little confusing.

mjzffr:
@jgraham Agreed. It's now affectionately called Listener.

PR #160, opened 2018-04-16
When a landing reaches central mark as complete and trigger the next one

jgraham:
Also, handle backouts of landings by ensuring that the relevant
landing is set to "open" and marked as having an error. Togehter this
means that we can't start a new landing until the present
one reaches central and any backout that happens before that will
prevent a new landing from starting.

### DIFF
--- a/sync/base.py
+++ b/sync/base.py
@@ -1,3 +1,4 @@
+import itertools
 import json
 import os
 import re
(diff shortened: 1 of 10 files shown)


sync/base.py
```diff
@@ -712,6 +713,59 @@ def __repr__(self):
                                self.sync_type,
                                self._process_name)
 
+    @classmethod
+    def prev_gecko_commit(cls, git_gecko, repository_name, base_rev=None, default=None):
+        """Get the last gecko commit processed by a sync process.
```
mjzffr: docstrings! 😍

PR #167, opened 2018-04-20
Don't push to try if a downstream sync has an error.

jgraham:
Always call update_commits before next_try_push. In update_commits,
unset the error flag if we are successful. Then if we are not
successful, we never end up with a try push (and we can explicitly
check for that condition to be sure). This prevents the case as with
PR 10536 where we have a merge conflict and then get some event that
causes a try push anyway.

### DIFF
--- a/sync/downstream.py
+++ b/sync/downstream.py
@@ -174,6 +174,11 @@ class DownstreamSync(base.SyncProcess):
         if not self.requires_try or self.status != "open":
             return
 
+        self.update_commits()
+
+        if self.error:
+            return
+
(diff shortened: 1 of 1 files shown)


sync/downstream.py
```diff
@@ -629,21 +631,29 @@ def try_push_complete(git_gecko, git_wpt, try_push, sync):
```
mjzffr: What about calling update_commits above Line 624?
Would it make sense to just add a call to update_commits in next_try_push instead?

mjzffr: It might be prudent to update test_downstream similarly.

PR #173, opened 2018-04-24
Refactor the taskcluster code to store tasks in an object.

jgraham:
This changes the design so that rather than a set of functions we end
up with a TaskGroup object that stores the underlying list of tasks
and any number of TaskGroupView objects that provide an interface for
examining some subset of the tasks.

The motivating factor here is that we want the is_complete function to
be able to cope with cases where the try push is "complete" in the
sense that no more tests will run (without some extra jobs being
triggered), but not all builds completed, so there are tasks still
marked as unscheduled. To tell the difference between an unscheduled
task that will eventually run, and one that will never run, we need to
look at the tasks it depends on; if their status is failure then any
unscheduled tasks that depend on them will never be scheduled, so the
task group can be considered complete.


mjzffr:
@jgraham Looks good so far overall. I'll be able to take a closer look tomorrow morning.

sync/trypush.py
```diff
@@ -433,46 +434,39 @@ def wpt_tasks(self, force_update=False):
...
+
+        self._data["tasks"] = tasks.tasks
+        return tasks
+
+    def wpt_tasks(self, force_update=False):
+        tasks = self.tasks(force_update)
+        wpt_tasks = tasks.view(tc.is_suite_fn("web-platform-tests"))
+        return wpt_tasks
+
+    def validate_tasks(self):
```
mjzffr: Previously, tasks were "validated" with every call to `wpt_tasks`. Now, a bunch of points at which we may have identified `infra_fail` are no longer in place (I see only one call to `validate_tasks`), so I'm concerned we'll miss an infra failure or label it much later in the sync process. Validating the tasks in `try_push_complete` makes sense, but have you looked at all the old calls to `wpt_tasks` to make sure we're not counting on updating the `infra_fail` flag at those points?

jgraham: I think it makes sense that we only validate in the entry point. We might be missing some validation in the landing code, but I think we want different semantics in that case anyway.

jgraham:
This is ready for more review, I think.

mjzffr:
@jgraham Looks like you added some new tests for tc requests. Anything else to take a closer look at?

jgraham:
Nothing specific, it was just that and some bug fixes vs the original commit.

PR #175, opened 2018-04-25
Reorder operations during landing

jgraham:


sync/command.py
```diff
@@ -218,7 +218,7 @@ def land_to_gecko():
 
     if current_landing and current_landing.latest_try_push:
         try_push = current_landing.latest_try_push
-        if try_push.status == "complete" or tc.is_complete(try_push.wpt_tasks(force_update=True)):
+        if try_push.status != "complete" and tc.is_complete(try_push.wpt_tasks(force_update=True)):
             if try_push.infra_fail:
                 land_to_gecko()
```
mjzffr: In terms of clarity, `land_to_gecko` and `landing.push_to_gecko()` sound like the same thing. Maybe rename `land_to_gecko` to something like `update_landing`?

PR #179, opened 2018-04-27
Improve retriggering of stuck PRs.

jgraham:


sync/handlers.py
```diff
@@ -190,6 +190,14 @@ def __call__(self, git_gecko, git_wpt, body):
             logger.debug("No try push for SHA1 %s" % (sha1,))
             return
 
+        # Enforce the invariant that the taskgroup id is not set until
+        # the decision task is complete. This allows us to determine if a
+        # try push should have the expected wpt tasks just by checking if
+        # this is set
+        if result == "pending":
```
mjzffr: It could also be "running" and "unscheduled"

* Maybe safer if we return if it's not exception, failed, completed.
* Would be nice to use `tc.PENDING`, etc., in this handler

sync/command.py
```diff
@@ -440,16 +441,31 @@ def do_landable(git_gecko, git_wpt, *args, **kwargs):
 
     if kwargs["all"] or kwargs["retrigger"]:
         unlandable = unlanded_with_type(git_gecko, git_wpt, wpt_head, prev_wpt_head)
-        status_map = {UnlandableType.ready: "ready",
+
+        status_map = {UnlandableType.ready: "Ready",
```
mjzffr: I have a slight preference for seeing status_map and detail_map defined next to the enums they are referring to, rather than in this command.

sync/downstream.py
```diff
@@ -97,22 +107,53 @@ def pr_status(self, value):
...
-        return latest_try_push.stability or not self.requires_stability_try
+            return DownstreamAction.try_push
+
+        if latest_try_push.status != "complete":
+            return DownstreamAction.wait_try
+        if latest_try_push.infra_fail:
+            if len(self.latest_busted_try_pushes()) > 5:
+                message = ("Too many busted try pushes. "
+                           "Check the try results for infrastructure issues.")
+                self.error = message
```
mjzffr: We don't want to comment on the bug here?

sync/downstream.py
```diff
@@ -612,28 +631,34 @@ def try_push_complete(git_gecko, git_wpt, try_push, sync):
...
-                # TODO: consider marking the push an error here so that we can't land without manual
-                # intervention
-
-    if try_push.stability and disabled:
-        logger.info("The following tests were disabled:\n%s" % "\n".join(disabled))
-        # TODO notify relevant people about test expectation changes, stability
-        env.bz.comment(sync.bug, ("The following tests were disabled "
-                                  "based on stability try push:\n %s" % "\n".join(disabled)))
+        if not try_push.validate_tasks():
+            if len(sync.latest_busted_try_pushes()) > 5:
```
mjzffr: If we call `next_action` here instead, we won't have to repeat the latest_busted_try_pushes logic.

sync/downstream.py
jgraham: So, the reason I don't do that is that we don't have any way to know if the comment was already made. We probably do want some notification on the bug though.

sync/downstream.py
jgraham: I'm not sure I see what you mean. `next_action` doesn't seem to contain this logic?

jgraham: Oh, sorry. So now I remember :) The problem is that next_action shouldn't be stateful, so it can't set the bug comment. But! I didn't actually finish fixing this right. I'll do that :)

mjzffr: However, next_action does do everything except set the bug comment, so could call next_action + comment here, no?

@jgraham Anyway, the new changes so far lgtm, but you're not done yet? Let me know.

jgraham: The idea is that you can call next_action multiple times, but we presumably don't end up in `try_push_complete` unless we actually had a try push, so you don't end up commenting multiple times. It's probably not perfect, but I think it's not totally wrong?

Anyway I think this PR is "complete" but there are others based on it.

PR #181, opened 2018-04-30
Ensure that the metadata commit is the tip of downstream sync

jgraham:
If we have a downstream sync and new commits are added, we currently
just add those commits after any existing metadata update commit. This
means we can break later invariants that there is only a single such
commit, or simply not apply the metadata during landing. Instead we
remove the metadata commit before updates, and reapply it later. Since
applying the upstream commits may fail, we need to store this pointer
in the sync data.

### DIFF
--- a/sync/downstream.py
+++ b/sync/downstream.py
@@ -15,6 +15,7 @@ import git
 
 import base
 import bugcomponents
+import gitutils
 import log
 import notify
 import trypush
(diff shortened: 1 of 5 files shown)


sync/downstream.py
```diff
@@ -394,28 +395,49 @@ def wpt_to_gecko_commits(self):
...
+                            dest_prefix=env.config["gecko"]["path"]["wpt"],
+                            msg_filter=self.message_filter,
+                            metadata=metadata)
+        except Exception:
+            raise
+        else:
+            if (metadata_commit is not None and
+                self.gecko_commits[-1].metadata.get("wpt-type") != "metadata"):
+                gitutils.cherry_pick(gecko_work, metadata_commit)
+            self.data["metadata-commit"] = None
```
mjzffr: If the cherrypick succeeds, don't we want self.data["metadata-commit"] to refer to the new commit?

jgraham: Maybe… in practice we only need this to be set if we are in the middle of an update requiring some manual fixup. Otherwise we can always just look at the last `gecko_commit` to get the metadata commit. And not caching it when we don't need the value avoids cache invalidation problems.

PR #188, opened 2018-05-17
Fix the handling of decision tasks.

jgraham:
The handling of decision tasks previously contained a few bugs,
notably that we were assuming the statuses were in the
taskcluster-queue set, but actually the pulse message we are using
here is from treeherder, which has two different kinds of status,
"state" which is just about whether the job is complete and "result"
which is about whether it succeeded, and a different set of
corresponding strings.

In addition the handling of failed decision tasks was lacking; we were
immediately failing and complaining in the bug, but not retriggering
anything. Now we attempt 5 retriggers before giving up.

### DIFF
--- a/sync/handlers.py
+++ b/sync/handlers.py
@@ -181,9 +181,17 @@ class PushHandler(Handler):
 
 
 class TaskHandler(Handler):
+    """Handler for the task associated with a decision task completing.
+
+    Note that this is using the Treeherder event type at
+    https://docs.taskcluster.net/reference/integrations/taskcluster-treeherder/references/events
+    rather than the TaskCluster event type, as this allows us to filter only
+    Gecko Decision Tasks."""
+
     def __call__(self, git_gecko, git_wpt, body):
         sha1 = body["origin"]["revision"]
         task_id = tc.normalize_task_id(body["taskId"])
+        state = body["state"]
         result = body["result"]
 
         try_push = trypush.TryPush.for_commit(git_gecko, sha1)
@@ -195,27 +203,33 @@ class TaskHandler(Handler):
         # the decision task is complete. This allows us to determine if a
         # try push should have the expected wpt tasks just by checking if
         # this is set
-        if result not in (tc.SUCCESS, tc.FAIL, tc.EXCEPTION):
-            logger.info("Decision task is not yet complete")
+        if state != "completed" or result == "superseded":
+            logger.info("Decision task is not yet complete, status %s" % result)
             return
 
-        logger.info("Setting taskgroup id for try push %r to %s" % (try_push, task_id))
-        try_push.taskgroup_id = task_id
-
-        failed_builds = len(try_push.failed_builds())
-        if result != tc.SUCCESS or failed_builds:
-            try_push.status = "complete"
-            try_push.infra_fail = True
+        # If we retrigger, we create a new taskgroup, with id equal to the new task_id.
+        # But the retriggered decision task itself is still in the original taskgroup
+        if result == "success":
+            logger.info("Setting taskgroup id for try push %r to %s" % (try_push, task_id))
+            try_push.taskgroup_id = task_id
+        elif result in ("fail", "exception"):
             sync = try_push.sync(git_gecko, git_wpt)
             message = ("Decision task got status %s for task %s%s" %
                        (result, sha1, " PR %s" % sync.pr if sync and sync.pr else ""))
-            if failed_builds:
-                message = "The try push includes build failures. %s" % message
             logger.error(message)
-            if sync and sync.bug:
-                # TODO this is commenting too frequently on bugs
-                env.bz.comment(sync.bug,
-                               "Try push failed: decision task returned error")
+            task = tc.get_task(task_id)
+            taskgroup = tc.TaskGroup(task["taskGroupId"])
+            if len(taskgroup.tasks.view(
+                    lambda x: x["metadata"]["name"] == "Gecko Decision Task")) > 5:
+                try_push.status = "complete"
+                try_push.infra_fail = True
+                if sync and sync.bug:
+                    # TODO this is commenting too frequently on bugs
+                    env.bz.comment(sync.bug,
+                                   "Try push failed: decision task %i returned error" % task_id)
+            else:
+                client = tc.TaskclusterClient()
+                client.retrigger(task_id)
 
 
 class TaskGroupHandler(Handler):
--- a/sync/downstream.py
+++ b/sync/downstream.py
@@ -765,7 +765,15 @@ def try_push_complete(git_gecko, git_wpt, try_push, sync):
                 sync.error = message
                 env.bz.comment(sync.bug, message)
                 try_push.status = "complete"
+                try_push.infra_fail = True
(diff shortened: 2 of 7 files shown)


sync/handlers.py
```diff
@@ -195,27 +203,29 @@ def __call__(self, git_gecko, git_wpt, body):
...
+            if len(taskgroup.tasks_by_id[task_id]) > 5:
+                try_push.status = "complete"
+                try_push.infra_fail = True
+                if sync and sync.bug:
+                    # TODO this is commenting too frequently on bugs
+                    env.bz.comment(sync.bug,
+                                   "Try push failed: decision task returned error")
+            else:
+                client = tc.TaskclusterClient()
+                client.retrigger(task_id)
```
mjzffr: By the way, I tried calling retrigger on a decision task just now and our client doesn't have sufficient scopes. (Retriggering other kinds of tasks works fine.) Then, lucky me, I ran into a bug when trying to add the missing scopes to our client: https://bugzilla.mozilla.org/show_bug.cgi?id=1462459

In any case, related to my previous comment, maybe `taskcluster.Queue.rerun` is more appropriate here anyway. It's especially for failed/exception tasks and I think it would keep the same task_id.

jgraham: `rerunTask` is marked as deprecated in the docs, so I'm a little worried about using it. Is that what you meant or is there something else?

sync/handlers.py
```diff
@@ -195,27 +203,40 @@ def __call__(self, git_gecko, git_wpt, body):
...
+                if sync and sync.bug:
+                    # TODO this is commenting too frequently on bugs
+                    env.bz.comment(sync.bug,
+                                   "Try push failed: decision task %i returned error" % task_id)
+            else:
+                # rerunTask is marked as deprecated, but
+                # <&jonasfj> hmm, jgraham I suspect I'll never be able to kill rerunTask,
+                #            so maybe just use it :)
+                client = tc.TaskclusterClient()
+                client.queue.rerunTask(task_id)
```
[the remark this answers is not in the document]
mjzffr: @jgraham I just read your discussion with jonas and I'm on the fence. 

I tested what the retriggering looks like: https://treeherder.mozilla.org/#/jobs?repo=try&revision=527702bdff5deaaa2d74cce1301ecd1717800db&selectedJob=179068982

In that try push, the 4th repetition for any job is the result of me retriggering the Decision Task via the python client. (So the retriggered Decision Task doesn't appear in Treeherder.) The jobs scheduled as a result of that decision task have a different task group id (their taskgroup id == retriggered Decision Task id, as expected). 

So every time we retrigger the Decision Task, we have to update the taskgroup id for the TryPush -- that doesn't sound too terrible. I think it really only needs to be updated in that one place and then everything else we have in place should still work.

If that turns out to be more complicated than I imagine, I guess rerunTask is the way to go.
What do you think?

sync/handlers.py
```diff
@@ -195,27 +203,34 @@ def __call__(self, git_gecko, git_wpt, body):
...
-        if result not in (tc.SUCCESS, tc.FAIL, tc.EXCEPTION):
-            logger.info("Decision task is not yet complete")
+        if state != "completed" or result == "superseded":
+            logger.info("Decision task is not yet complete, status %s" % result)
             return
 
-        logger.info("Setting taskgroup id for try push %r to %s" % (try_push, task_id))
-        try_push.taskgroup_id = task_id
-
-        failed_builds = len(try_push.failed_builds())
```
mjzffr: So I think the `failed_builds` logic should be moved to `TaskGroupHandler`.

sync/handlers.py
```diff
@@ -195,27 +203,34 @@ def __call__(self, git_gecko, git_wpt, body):
...
-        if result != tc.SUCCESS or failed_builds:
-            try_push.status = "complete"
-            try_push.infra_fail = True
+        # If we retrigger, we create a new taskgroup, with id equal to the new task_id.
+        # But the retriggered decision task itself is still in the original taskgroup
+        # In order to count the number of retriggers we have, we always retrigger the
+        # first scheduled decision task, so that all decision tasks are in a single taskgroup
```
mjzffr: That's not what we're actually doing here,  since `task_id` doesn't always represent the original decision task. We're actually retriggering the most recent retriggered decision task, which is fine. 

Unfortunately, `body` doesn't give us the task group id, so I think we need to save some state. Maybe a retries count in TryPush, rather than saving the original taskgroup id? In that case, we can also modify `retrigger` to make the taskgroup id match the task id whenever we're retry a decision task, thus maintaining the convention that decision task id and taskgroup id always match.

mjzffr: I mean, in `retrigger` we can set the payload to have whatever taskgroup id we like...

jgraham: So I think the comment could be wrong, but the logic is "right", depending on what properties we want to preserve.

If we retrigger the `task_id` we are given then the `taskGroupId` will be the same in the retriggered task. For the first retrigger, that will be the same as the original `taskGroupId` and by induction it will continue to be the same for all subsequent retriggers. That's what's implemented here and it does make the logic simple because we have a single taskgroup with all the decision tasks, so it's obvious how many times we have retriggered, just by counting.

It is also possible to set the `taskGroupId` to be equal to the `task_id`, in which case we preserve the property that they are identical for the decision task. *Not* doing that, I think, only has the consequence that the decision task itself ends up in a different task group from the other jobs. However if we do that there's no easy way to look up all the retriggers (I guess we could use treeherder's API if we really wanted), so we probably have to store a retrigger count on the `TryPush` object. 

I think the implemented approah has the benefit of being the simplest thing that could possibly work, but if you want to preserve the stronger invariant on decision task's ids matching, we need to do something more complex.

mjzffr: @jgraham I agree with most of that, except that the logic is still wrong. :) The handler is not receiving a task's taskgroup id in `body`, so although all the retriggered decision tasks do belong to the same taskgroup, the following code is not actually getting that taskgroup:

```
# task_id is the id of the retriggered decision task, not the original taskgroup
taskgroup = tc.TaskGroup(task_id)
if len(taskgroup.tasks.view(lambda x: x["metadata"]["name"] == "Gecko Decision Task")) > 5:
    ...
```
The response we'll get is something like https://queue.taskcluster.net/v1/task-group/BpxdADiRTnu67LXP0OYdow/list -- it won't include any decision tasks at all because it's the list of tasks scheduled as a result of the retriggered decision task.

Now that I think about it, you can make an extra request to get more info about the retriggered task: e.g. `tc.get_task(task_id).get("taskGroupId")` and then use that make your `tc.TaskGroup`.

sync/handlers.py
```diff
@@ -181,9 +181,17 @@ def __call__(self, git_gecko, git_wpt, body):
 
 
 class TaskHandler(Handler):
+    """Handler for the task associated with a decision task completing.
+
+    Note that this is using the Treeherder event type at
```
mjzffr: @jgraham I just realized that the filter in listen.py must be wrong too: https://github.com/mozilla/wpt-sync/blob/master/sync/listen.py#L194

How is it even working at all? I think we want it to check that `body["state"]` is "completed".

PR #190, opened 2018-05-21
Handle try: in downstream commit message.

jgraham:
Gecko's hg hooks reject commit messages with try: in them. It turns out that Chrome uses
similar syntax but allows it in commits, so we end up with this case being hit
semi-regularly.

### DIFF
--- a/sync/downstream.py
+++ b/sync/downstream.py
@@ -558,12 +558,20 @@ class DownstreamSync(base.SyncProcess):
         return commits
 
     def message_filter(self, msg):
+        # It turns out that the string "try:" is forbidden anywhere in gecko commits,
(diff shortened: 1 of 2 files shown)


sync/downstream.py
```diff
@@ -558,12 +558,20 @@ def unlanded_commits_same_files(self):
         return commits
 
     def message_filter(self, msg):
+        # It turns out that the string "try:" is forbidden anywhere in gecko commits,
+        # because we (mistakenly) think that this always means it's a try string. So we insert
+        # a ZWSP which means that the try syntax regexp doesn't match, but the printable
```
mjzffr: Ha, excellent!

PR #201, opened 2018-06-05
Fix retrieving TaskGroupView

mjzffr:


mjzffr:
Doh that I didn't catch it in review last time. :)

PR #203, opened 2018-07-04
On landing, do a try push with metadata updates but without retrggering.

jgraham:
The current design was predicated on the idea that metadata from PR
try runs would usually be accurate enough that metadata updates
wouldn't be required on landing. But in practice this doesn't seem to
be the case. Given that, the disadvantage of the previous approach is
that end up retriggering many hundreds of jobs on each push, and the
resulting metadata update causes problems like poor performance and
OOM due to the number of logs being consumed.

A simple switcharound implemented here is, assuming the first try push
fails, to do a metadata update without any retriggers and perform a
new try push. If that try push also fails, we fall back to the current
approach of retriggering failed jobs to check for intermittents.


andreastt:
I can give an r+ on the grounds that it looks technically sound, but I don’t really feel comfortable reviewing htis.

PR #210, opened 2018-09-06
Improve some logging

jgraham:


sync/landing.py
```diff
@@ -606,7 +607,8 @@ def push(landing):
             raise AbortError(err)
 
         if old_head == landing.gecko_commits.head.sha1:
-            err = "Landing push failed and rebase didn't change head"
+            err = ("Landing push failed and rebase didn't change head:%s" %
+                   ("\n%s" % err if err else ""))
```
whimboo: Using two levels of formatting is weird. Can you please remove it?

jgraham: Not easily? I mean I can use an intermediate variable but it doesn't seem worthwhile.

whimboo: I meant why not just:

```
"Landing push failed and rebase didn't change head:\n%s" % err if err else ""
```

PR #215, opened 2018-09-06
Ensure that we delete the upstream branch for a sync once it's merged

jgraham:


sync/upstream.py
```diff
@@ -425,14 +425,22 @@ def try_land_pr(self):
...
         return False
 
+    def finish(self, status="complete"):
+        super(UpstreamSync, self).finish(status)
+        if status in ("wpt-merged", "complete") and self.remote_branch:
+            # Delete the remote branch after a merge
+            try:
+                self.git_wpt.remotes.origin.push(self.remote_branch, delete=True)
+            except git.GitCommandError:
+                pass
```
whimboo: Worth a log entry? Remotes could add up if something is fundamentally broken. Or how are remaining remote branches tracked? At least those aren't getting reset to `None` here.

PR #218, opened 2018-09-06
Attempt to retrigger failed tasks after pushing to gecko

jgraham:


jgraham:
Oh this is erroring because there's no queue running so it hangs. Need to fix that in the code anyway.

jgraham:
@dhdavvie we should ressurect this; it tries to retrigger failed jobs after we do a gecko push because that update means that things which previously failed to apply might now apply correctly. But the patch failed if there wasn't a queue running which can be the case in interactive use or in tests. Maybe you could look at if there's some way to handle that?

dhdavvie:
So this is looking interesting as none of the standard timeout options seem to work, and I am not able to reliably see something that can be used identify the presence (or lack thereof) of a queue. Here is my current thought process:

The `amqp` object has a list of queues, which would be marked as unbound if nothing is there in a `_is_bound` property. During testing it looks like all of these are unbound so can be used to skip the retriggering of tasks. If we are sure about which Queue (is it just the 'celery' queue?) would need to be checked in prod, then this `_is_bound` check could be used. Otherwise I am not seeing another way.

dhdavvie:
This has been revisited in a new PR.

PR #227, opened 2018-10-16
Don't send patches through a UTF8 encode/decode step

jgraham:
This will garble any commits that are testing non-UTF8 encodings.

### DIFF
--- a/sync/commit.py
+++ b/sync/commit.py
@@ -160,7 +160,11 @@ class Commit(object):
             args = ("--", prefix)
         else:
             args = ()
-        return self.repo.git.show(self.sha1, format="", patch=True, *args).strip() == ""
+        return self.repo.git.show(self.sha1,
+                                  format="",
+                                  patch=True,
+                                  stdout_as_string=False,
+                                  *args).strip() == ""
 
     def tags(self):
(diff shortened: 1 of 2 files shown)


sync/commit.py
```diff
@@ -160,7 +160,11 @@ def is_empty(self, prefix=None):
             args = ("--", prefix)
         else:
             args = ()
-        return self.repo.git.show(self.sha1, format="", patch=True, *args).strip() == ""
+        return self.repo.git.show(self.sha1,
+                                  format="",
+                                  patch=True,
+                                  stdout_as_string=False,
```
whimboo: `False` is the default so you can keep the line untouched.

jgraham: No, `True` is the default, that's the bug that's being fixed here

whimboo: So what does this line do then? https://github.com/mozilla/wpt-sync/blob/c50d5d75743355c403bead18391a8f8040b85888/sync/commit.py#L187

jgraham: That code isn't being called in this case.

whimboo: So this shows my missing knowledge of this code as pointed out 27 days ago. Given that I don't have the time to dive into all this, I trust you and the passing new tests. Lets get it shipped.

PR #229, opened 2018-10-16
Go thorugh update landing code when the landing appears ready.

jgraham:
This ensures that we have a single entry point and run any logic about retriggering etc.


sync/command.py
```diff
@@ -268,10 +268,7 @@ def update_landing():
                                           allow_push=kwargs["push"],
                                           accept_failures=accept_failures)
         elif try_push.status == "complete" and not try_push.infra_fail:
-            landing.push_to_gecko(git_gecko,
-                                  git_wpt,
-                                  current_landing,
-                                  allow_push=kwargs["push"])
+            update_landing()
```
whimboo: Can we have a test for this change so that I can better understand which state this is covering?

PR #382, opened 2019-07-17
Download fixes

jgraham:


sync/tc.py
```diff
@@ -220,32 +225,63 @@ def download_logs(self, destination, file_names, retry=5):
...
+    try:
+        artifacts = fetch_json(artifacts_base_url, session=session)
+    except requests.HTTPError as e:
+        logger.warning(e.message)
+    artifact_urls = ["%s/%s" % (artifacts_base_url, item["name"])
+                     for item in artifacts["artifacts"]
+                     if any(item["name"].endswith("/" + file_name)
+                            for file_name in file_names)]
+
+    run = status["runs"][-1]
```
mjzffr: I'm assuming this works, but it's notable to me that now we're just looking at one run instead of a list of runs. I guess there's always just one run per task and a retriggered job is in an independent task? I don't remember the structure of the task status. Pointing it out in case there was an oversight.

jgraham: Yeah, so a retrigger is not a rerun, it's a different task. reruns are mostly the automatic restarts in cases where the spot instance is discarded or whatever. I think the assumption that the final run is always the useful one is pretty reasonable (the fetchlogs tool assumes this and I haven't noticed it causing a problem yet, so there's also some empirical evidence that the theory works). The advantage of not looking at all the runs is obviously that it's a bit faster.

PR #383, opened 2019-07-17
Lookup upstream syncs by PR id using an index

jgraham:


sync/index.py
```diff
@@ -317,25 +340,30 @@ class PrIdIndex(Index):
...
-            logger.warning("Corrupt process %s" % item)
+                cls = DownstreamSync
+            elif process_name.subtype == "upstream":
+                cls = UpstreamSync
+            if not cls:
+                continue
+            try:
+                sync = cls(git_gecko, git_wpt, process_name)
+            except ValueError:
+                import pdb
```
mjzffr: A stray pdb.

PR #400, opened 2019-07-24
Support adding needinfo when upstream PR can't merge

jgraham:
This encourages authors to investigate when their PRs fail to merge
which should avoid some of the problem where we have many PRs that are
not merged fo rtrivial failures or need to manually ping authors.

This adds the beginnings of a new API for working with bugs that
allows making multiple changes before updating bugzilla, so that the
needinfo request and the comment come from a single conceptual place.


jgraham:
Updated to fix the review comments.

Issue #379, opened 2019-07-16
Don't try to reland upstream PRs if the commits didn't change


jgraham:
When we upstream a change from gecko to wpt and then try to update the gecko sync point past that PR we try to reapply the changes from the PR in case someone pushed additional changes not present in the gecko branch. But in the common case where this doesn't happen we should be able to tell because the head of the PR will be the same as the last commit that the sync pushed. In this case we should be able to skip relanding the PR altogether. This will also allow us to remove the current hack where we skip in cases that the files in GitHub and m-c look identical.

dhdavvie:
Would this be fixed by checking the commit hashes in [gecko_push](https://github.com/mozilla/wpt-sync/blob/master/sync/upstream.py#L810)? Something along the lines of:

```py
if rev == last_sync_point.commit.sha1:
    logger.info("Last sync point is the same as this revision")
    return
```

And place this right before the lock? Or as an elif to the [ancestor check](https://github.com/mozilla/wpt-sync/blob/master/sync/upstream.py#L816)?

PR #403, opened 2019-07-25
Don't try to reland upstream PRs if the commits didn't change.

dhdavvie:
#379 

Added a test that replicates a potential scenario where this would arise:
1 - Create a pull request to wpt that is unrelated to Gecko and merge it
2 - Create a Gecko bug, upstream it, and merge this PR into wpt unchanged.
3 - Land the wpt changes into try/central

The LandingSync returned at step 3 should only have the PR from step 1 going into try, as it would have identified step 2's commit being from a Gecko bug and not even bothered trying to apply it.

### DIFF
--- a/sync/landing.py
+++ b/sync/landing.py
@@ -233,24 +233,15 @@ Automatic update from web-platform-tests\n%s
 
         logger.info("Upstream files changed:\n%s" % "\n".join(sorted(upstream_changed)))
 
-        if isinstance(sync, upstream.UpstreamSync):
-            remote_paths = list(upstream_changed)
-            local_paths = [os.path.join(env.config["gecko"]["path"]["wpt"], path)
-                           for path in upstream_changed]
-            has_changes = False
-            for local_path, remote_path in zip(local_paths, remote_paths):
-                try:
-                    if (self.git_wpt.git.show("origin/master:%s" % remote_path) !=
-                        git_work_gecko.git.show("HEAD:%s" % local_path)):
-                        logger.info("Changes in %s" % local_path)
-                        has_changes = True
-                        break
-                except git.GitCommandError:
-                    has_changes = True
-                    break
-            if not has_changes:
-                logger.info("Upstream sync doesn't introduce any gecko changes")
-                return
+        # If this is originally an UpstreamSync and no new changes were introduced to the GH PR
+        # then we can safely skip and not need to re-apply these changes. Compare the hash of
+        # the upstreamed gecko commits against the final hash in the PR.
+        last_gecko_rev = wpt_commits[-1].metadata.get('gecko-commit', None)
+        if isinstance(sync, upstream.UpstreamSync) and \
+                sync.upstreamed_gecko_commits[-1].canonical_rev == last_gecko_rev:
+
+            logger.info("Upstream sync doesn't introduce any gecko changes")
+            return
 
         if copy:
             commit = self.copy_pr(git_work_gecko, git_work_wpt, pr, wpt_commits,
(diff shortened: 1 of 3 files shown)


sync/landing.py
```diff
@@ -233,24 +233,15 @@ def add_pr(self, pr_id, sync, wpt_commits, copy=True, prev_wpt_head=None):
...
-                    break
-            if not has_changes:
-                logger.info("Upstream sync doesn't introduce any gecko changes")
-                return
+        # If this is originally an UpstreamSync and no new changes were introduced to the GH PR
+        # then we can safely skip and not need to re-apply these changes. Compare the hash of
+        # the upstreamed gecko commits against the final hash in the PR.
+        last_gecko_rev = wpt_commits[-1].metadata.get('gecko-commit', None)
+        if isinstance(sync, upstream.UpstreamSync) and \
+                sync.upstreamed_gecko_commits[-1].canonical_rev == last_gecko_rev:
```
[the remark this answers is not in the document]
dhdavvie: Unfortunately not as they don't seem to match (although could be a side-effect of the spoofing from the test). However `sync._upstreamed_gecko_head == wpt_commits[-1].sha1` could work. Does `sync._upstreamed_gecko_head` change if someone pushes/squashes upstream?

What is the preference on using `wpt_commits` vs `sync.wpt_commits`?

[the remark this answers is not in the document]
jgraham: Oh yes, upstreamed gecko commits is just wrong. What should work is comparing `sync.wpt_commits.head.sha1` with the sha1 of the commit at the `refs/remotes/origin/pr/{sync.pr_id}` branch. The latter doesn't seem to have a simple accessor so I think you probably want to add a `pr_commits` property or something.

sync/landing.py
```diff
@@ -233,24 +233,15 @@ def add_pr(self, pr_id, sync, wpt_commits, copy=True, prev_wpt_head=None):
...
-                    break
-            if not has_changes:
-                logger.info("Upstream sync doesn't introduce any gecko changes")
-                return
+        # If this is originally an UpstreamSync and no new changes were introduced to the GH PR
+        # then we can safely skip and not need to re-apply these changes. Compare the hash of
+        # the upstreamed gecko commits against the final hash in the PR.
+        pr_ref = 'origin/pr/{}'.format(pr_id)
+        if (isinstance(sync, upstream.UpstreamSync) and
+                sync.wpt_commits.head.sha1 == sync.git_wpt.refs[pr_ref].commit.hexsha):
```
jgraham: I think this is logically correct, but can we create a `pr_commits` property on  `UpstreamSync` that returns thse commits and handles cases like the sync not having a PR? 

Also the indentation of the second line looks incorrect.

dhdavvie: That sounds like a good idea, I will work on implementating that.

Its done to avoid it being confused as the next logical line. But obviously can move it back an indent to stick with the styling of the rest of the codebase.

sync/upstream.py
```diff
@@ -469,6 +469,29 @@ def finish(self, status="complete"):
...
+        """
+        if not self.pr:
+            logger.error("No PR ID found for %s" % self.process_name)
+            return
+
+        pr_ref = 'origin/pr/{}'.format(self.pr)
+
+        if pr_ref not in self.git_wpt.refs:
+            # Maybe we need to fetch?
+            self.git_wpt.remotes.origin.fetch()
```
jgraham: I think it's OK to just error here. In general we make sure we fetch before running any task rather than having random code fetch.

dhdavvie: makes sense, fixed

PR #408, opened 2019-07-31
Added a `pr_commits` property to UpstreamSync

dhdavvie:
[WIP]

This property is to get the commits of a WPT PR that is related to a Gecko Bug. It tries to use the local wpt git repository as much as possible, and falls back to using the GitHub API if something is inconsistent.

### DIFF
--- a/sync/upstream.py
+++ b/sync/upstream.py
@@ -9,7 +9,7 @@ from mozautomation import commitparser
 
 import log
 import commit as sync_commit
-from base import entry_point, BranchRefObject
+from base import entry_point
 from downstream import DownstreamSync
 from errors import AbortError
 from env import Environment
--- a/sync/upstream.py
+++ b/sync/upstream.py
@@ -14,6 +14,7 @@ from downstream import DownstreamSync
 from errors import AbortError
 from env import Environment
 from gitutils import update_repositories, gecko_repo
+from gh import AttrDict
 from lock import SyncLock, constructor, mut
 from sync import CommitFilter, LandableStatus, SyncProcess, CommitRange
 from repos import pygit2_get
@@ -482,11 +483,9 @@ class UpstreamSync(SyncProcess):
         if not pr_head_reachable:
             merge_base = self.git_wpt.merge_base(origin_master_sha, pr_head.sha1)
         else:
-
             if not self.merge_sha:
                 raise ValueError('The merge SHA for %s could not be found in the UpstreamSync' %
                                  self.process_name)
-
             merge_commit = sync_commit.WptCommit(self.git_wpt, self.merge_sha)
 
             # If the commit has two parents, one of them being our pr head, it is a merge commit
@@ -507,7 +506,7 @@ class UpstreamSync(SyncProcess):
 
         # Create a CommitRange object and return it
         base = sync_commit.WptCommit(self.git_wpt, merge_base)
-        head_ref = BranchRefObject(self.git_wpt, self.process_name, sync_commit.WptCommit)
+        head_ref = AttrDict({'commit': pr_head})
         return CommitRange(self.git_wpt, base, head_ref, sync_commit.WptCommit, CommitFilter())
 
 
--- a/sync/upstream.py
+++ b/sync/upstream.py
@@ -472,24 +472,31 @@ class UpstreamSync(SyncProcess):
             )
             return
 
-        # Check that the merge base we have is valid
-        # TODO this also passes if the PR was rebased to a commit after our recorded merge base
-        recorded_merge_base = self.wpt_commits.base.sha1
-        merge_base_check = self.git_wpt.merge_base(pr_head, recorded_merge_base)
-        if len(merge_base_check) == 0 or merge_base_check[0].hexsha != recorded_merge_base:
-            # TODO Merge bases are inconsistent, what to do?
-            return
+        # pr = env.gh_wpt.get_pull(self.pr)
+        tag = 'merge_pr_%d' % self.pr
+        if tag not in self.git_wpt.tags:
+            logger.error("Merge Tag not found for %s" % tag)
+        merge_commit_sha = self.git_wpt.tags[tag].commit
+
+        # If the merge commit is not the pr head nor a descendent then it was squash merged
+        if merge_commit_sha != pr_head and not self.git_wpt.is_ancestor(pr_head, merge_commit_sha):
+            merge_base = self.git_wpt.merge_base(merge_commit_sha, pr_head)
+            if len(merge_base) == 0:
+                logger.error('Merge base could not be determined')
+                return
+            else:
+                merge_base = merge_base[0]
 
         # Add all the commits ref Head -> merge base to a list, these should be the PR commits
         head_commit = self.git_wpt.commit(rev=pr_head)
         commits = [head_commit]
         for parent in head_commit.iter_parents():
-            if parent == merge_base_check[0]:
+            if parent == merge_base:
                 break
             commits.append(parent)
 
         # Reverse the list so the first commit is the oldest
-        return reversed(commits)
+        return reversed(commits)  # TODO: Should this be a CommitRange?
 
 
 def commit_message_filter(msg):
--- a/sync/upstream.py
+++ b/sync/upstream.py
@@ -466,70 +466,52 @@ class UpstreamSync(SyncProcess):
     def pr_commits(self):
         pr_head = self.pr_head
         if not pr_head:
-            logger.error(
-                "Can't get PR commits as the ref head could not be found for %s" %
-                self.process_name
-            )
-            return
+            raise ValueError("Can't get PR commits as the ref head could not be found for %s" %
+                             self.process_name)
         else:
-            pr_head = self.git_wpt.commit(pr_head)
+            pr_head = sync_commit.WptCommit(self.git_wpt, pr_head)
 
-        pr = None
+        merge_base = []
 
-        # Retrieve the commit that pushed these changes to master
-        tag = 'merge_pr_%d' % self.pr
-        if tag in self.git_wpt.tags:
-            merge_commit_sha = self.git_wpt.tags[tag].commit
-        elif self.status in ('wpt-merged', 'complete'):
-            logger.warning("Merge Tag not found for %s" % tag)
-            pr = env.gh_wpt.get_pull(self.pr)
-            merge_commit_sha = pr["merge_commit_sha"]
-        elif self.pr_status == 'open':
-            # PR must still be open, just use master as our 'merge commit'
-            merge_commit_sha = self.git_wpt.head.commit.hexsha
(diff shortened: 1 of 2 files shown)


sync/upstream.py
```diff
@@ -462,6 +462,83 @@ def pr_head(self):
         ref = self.git_wpt.refs[pr_ref]
         return ref.commit.hexsha
 
+    @property
+    def pr_commits(self):
+        pr_head = self.pr_head
+        if not pr_head:
+            logger.error(
```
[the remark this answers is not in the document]
dhdavvie: Thanks, I will implement a ValueError here.

sync/upstream.py
```diff
@@ -462,6 +462,83 @@ def pr_head(self):
...
+                "Can't get PR commits as the ref head could not be found for %s" %
+                self.process_name
+            )
+            return
+        else:
+            pr_head = self.git_wpt.commit(pr_head)
+
+        pr = None
+
+        # Retrieve the commit that pushed these changes to master
```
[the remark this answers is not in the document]
dhdavvie: I agree, looking at it now it looks overly complex. I will revisit the logic with the suggestions you made.

sync/upstream.py
```diff
@@ -462,6 +462,83 @@ def pr_head(self):
...
+            # Singular parent, fast forward merge
+            elif len(merge_commit.parents) == 1 and merge_commit == pr_head:
+
+                # Fall back to the API for this
+                if not pr:
+                    pr = env.gh_wpt.get_pull(self.pr)
+                merge_base = self.git_wpt.commit(pr["base"]["sha"])
+
+        # Squash and Merge
+        # If the pr head is not reachable from master then it must be a typical squash merge
```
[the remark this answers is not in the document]
dhdavvie: Oh I didn't realise this could occur, I was under the impression the entire PR would rebase and move to M2 as the merge base. Thanks for pointing this out.

sync/upstream.py
```diff
@@ -462,6 +462,65 @@ def pr_head(self):
...
+        merge_base = []
+
+        # Check if the PR Head is reachable from origin/master
+        origin_master_sha = self.git_wpt.refs['origin/master'].commit.hexsha
+        pr_head_reachable = self.git_wpt.is_ancestor(pr_head.sha1, origin_master_sha)
+
+        # If not reachable, then it either hasn't landed yet, it was a Squash + Merge,
+        # or a Rebase and merge.
+        if not pr_head_reachable:
+            # TODO: how to effectively detect Rebase and Merge?
```
dhdavvie: I'm still a bit unsure of how to detect this scenario, other than checking this merge base against the one returned by GH, but even then I am unsure which merge base it would give (gut feeling says the new merge base)

sync/upstream.py
```diff
@@ -466,78 +466,60 @@ def pr_head(self):
...
-        elif self.status in ('wpt-merged', 'complete'):
-            logger.warning("Merge Tag not found for %s" % tag)
-            pr = env.gh_wpt.get_pull(self.pr)
-            merge_commit_sha = pr["merge_commit_sha"]
-        elif self.pr_status == 'open':
-            # PR must still be open, just use master as our 'merge commit'
-            merge_commit_sha = self.git_wpt.head.commit.hexsha
+        # Check if the PR Head is reachable from origin/master
+        origin_master_sha = self.git_wpt.refs['origin/master'].commit.hexsha
+        pr_head_reachable = self.git_wpt.is_ancestor(pr_head.sha1, origin_master_sha)
```
jgraham: You can just use `origin/master` directly here rather than a sha

sync/upstream.py
```diff
@@ -466,78 +466,60 @@ def pr_head(self):
...
-            )
-            return
+            raise ValueError("Can't get PR commits as the ref head could not be found for %s" %
+                             self.process_name)
         else:
-            pr_head = self.git_wpt.commit(pr_head)
+            pr_head = sync_commit.WptCommit(self.git_wpt, pr_head)
 
-        pr = None
+        merge_base = []
```
jgraham: I don't know why this is a list since it's either 0 or 1 commits

sync/upstream.py
jgraham: We don't need to detect how the actual merge was done as such, we only need to care about the final state of the tree, so I'm not sure exactly what the concern is here.

sync/upstream.py
```diff
@@ -466,78 +466,60 @@ def pr_head(self):
...
+            # Singular parent and pr head is the merge commit, it is a fast forward merge
+            elif len(merge_commit.commit.parents) == 1 and merge_commit.commit == pr_head.commit:
+                merge_base = [self.wpt_commits.base.commit]
 
         # Check for inconsistency in the merge base
-        if merge_base.hexsha != self.wpt_commits._base_sha:
-            # If we are inconsistent, fall back to the GH API
-            if not pr:
-                pr = env.gh_wpt.get_pull(self.pr)
+        if len(merge_base) == 0 or merge_base[0].hexsha != self.wpt_commits.base.sha1:
```
jgraham: I think we can drop this consistency check; as long as the `merge_sha` exists we have the latest data from github.

sync/upstream.py
```diff
@@ -466,78 +466,60 @@ def pr_head(self):
...
-        else:
-            merge_base = self.git_wpt.merge_base(merge_commit_sha, pr_head)
-            if len(merge_base) == 0:
-                logger.error('Merge base could not be determined')
-                return
-            else:
-                merge_base = merge_base[0]
+            # Singular parent and pr head is the merge commit, it is a fast forward merge
+            elif len(merge_commit.commit.parents) == 1 and merge_commit.commit == pr_head.commit:
+                merge_base = [self.wpt_commits.base.commit]
```
jgraham: I guess this is the same as `self.data['wpt-base']`, but I think it's a little clearer to use the latter.

sync/upstream.py
```diff
@@ -466,78 +466,60 @@ def pr_head(self):
...
-        # If the pr head is not reachable from master then it must be a typical squash merge
-        else:
-            merge_base = self.git_wpt.merge_base(merge_commit_sha, pr_head)
-            if len(merge_base) == 0:
-                logger.error('Merge base could not be determined')
-                return
-            else:
-                merge_base = merge_base[0]
+            # Singular parent and pr head is the merge commit, it is a fast forward merge
+            elif len(merge_commit.commit.parents) == 1 and merge_commit.commit == pr_head.commit:
```
jgraham: This set of if conditions isn't exhaustive.

sync/upstream.py
```diff
@@ -466,78 +466,60 @@ def pr_head(self):
...
+            pr = env.gh_wpt.get_pull(self.pr)
             merge_base = self.git_wpt.commit(pr["base"]["sha"])
+            self.wpt_commits.base = merge_base.hexsha
+        else:
+            merge_base = merge_base[0]
 
         # Add all the commits ref Head -> merge base to a list, these should be the PR commits
-        head_commit = self.git_wpt.commit(rev=pr_head)
+        head_commit = self.git_wpt.commit(rev=pr_head.sha1)
         commits = [head_commit]
```
jgraham: This is something we can do with GitPython directly; get it to give us an iterator over all the commits in the range 'merge_base..head`; see `CommitRange` for example (it would be nice to use that directly in fact).

sync/upstream.py
dhdavvie: I guess I was concerned about if this is used to determine which changes to apply to central, we may end up applying the same changes twice, but yes for the purposes of finding the commits in the PR it has no relevance.

sync/upstream.py
dhdavvie: The merge_base method returns a list with either 0 and 1 commit, and since I use it twice I create an empty list here so that later on I can do a single `if len(merge_base) == 0` check instead of checking the length each call.

sync/upstream.py
dhdavvie: It seems as though `self.data['wpt-base']` just stores the base branch, and not the base commit.

sync/upstream.py
dhdavvie: Could change this to an else, so that if we don't think its a merge commit, we always use our stored merge base?

sync/upstream.py
dhdavvie: Ahh yes I had a feeling CommitRange would be the more appropriate tool here, thanks!

sync/upstream.py
jgraham: It's changed to the base commit after we merge it, exactly for this reason.

sync/upstream.py
jgraham: OK, it's a bit strange, but I guess it makes sense.

sync/upstream.py
dhdavvie: Ahh I see, thanks

sync/upstream.py
jgraham: If we have a situation where we have PR commits like `[from gecko, from gecko, fixup]` and we land with a rebase we can obviously just ignore the first two when we come to do the landing. If they're landed with a squash, we can probably compare `pr_commits` with `wpt_commits`, notice that there's one new commit and just apply that. If there's something like `[from gecko but modified]` we probably just have to stick with the current approach of applying everything and hoping there aren't conflicts. But that's far fewer cases.

sync/upstream.py
jgraham: I think that makes sense in this case. In general it's better to explictly error in cases we can't handle rather than just letting the logic fail later. But in this case we can handle it, so…

sync/upstream.py
```diff
@@ -462,6 +462,54 @@ def pr_head(self):
...
+        # If not reachable, then it either hasn't landed yet, it was a Squash + Merge,
+        # or a Rebase and merge.
+        if not pr_head_reachable:
+            merge_base = self.git_wpt.merge_base(origin_master_sha, pr_head.sha1)
+        else:
+
+            if not self.merge_sha:
+                raise ValueError('The merge SHA for %s could not be found in the UpstreamSync' %
+                                 self.process_name)
+
```
jgraham: No blank line.

sync/upstream.py
```diff
@@ -462,6 +462,54 @@ def pr_head(self):
...
+        # Check that we found the merge base
+        if len(merge_base) == 0:
+            raise ValueError("Problem determining merge base for %s" % self.process_name)
+        else:
+            merge_base = merge_base[0]
+
+        # Create a CommitRange object and return it
+        base = sync_commit.WptCommit(self.git_wpt, merge_base)
+        head_ref = BranchRefObject(self.git_wpt, self.process_name, sync_commit.WptCommit)
```
jgraham: But the head_ref for the PR should be the `refs/prs/{}/head` or whatever it is that we use to get `pr_head`.

dhdavvie: yes of course, brain skipped a beat there

PR #412, opened 2019-08-01
Needinfo a config-defined list of users when bugs fail to apply to up…

jgraham:
…stream.

This prevents the case where a patch won't apply to upstream and we fail to notice for some time
because people forget to check the list of syncs in the error state. The list of users is intended
to be the people who have access to the prod sync instance and so can fix the problem (it would be
more convenient to notify a single shared account, but it's unclear how that ought to work).

### DIFF
--- a/config/dev/sync.ini
+++ b/config/dev/sync.ini
@@ -30,6 +30,7 @@ path.meta = testing/web-platform/meta
 worktree.max-count = 10
 logs.max-count = 20
 try.max-tests = 500
+upstream.needinfo-users=example@example.org,
 
 [web-platform-tests]
 repo.url = %ROOT%/remotes/web-platform-tests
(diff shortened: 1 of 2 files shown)


mjzffr:
@jgraham It it possible to configure the wptsync@mozilla.bugs account to send need-info email to wptsync@mozilla.com, to which we're all signed up?

jgraham:
> @jgraham It it possible to configure the [wptsync@mozilla.bugs](mailto:wptsync@mozilla.bugs) account to send need-info email to [wptsync@mozilla.com](mailto:wptsync@mozilla.com), to which we're all signed up?

That sounds like a good solution, although it does mean that the needinfos wouldn't appear in our personal lists (afaik). Maybe @dklawren can answer the question of whether it's possible.

dklawren:
You do not necessarily need to have the email sent to 
wptsync@mozilla.com. Anyone who wants to see the same notifications
as the wptsync@mozilla.bugs account can just *watch* that email by 
setting it in their preferences.

https://bugzilla.mozilla.org/userprefs.cgi?tab=email

David Lawrence
dkl@mozilla.com

------ Original Message ------

Issue #378, opened 2019-07-16
Add GitHub labels for PRs that are blocked in Gecko


jgraham:
e.g. ones that have a merge conflict that needs to be sorted. This would provide a convenient frontend that doesn't require going via the command line just to find out if there are issues.

dhdavvie:
May need to revisit this if we decide the labelling is not working as intended

PR #422, opened 2019-08-08
Add a github label when there is a downstream error

dhdavvie:
#378 

This will add a label to the GH PR if something goes wrong with the downstream sync process and an error is stored. Technically this error property can be set by something other than an error happening in Gecko so maybe a different name for the label is needed?

### DIFF
--- a/sync/downstream.py
+++ b/sync/downstream.py
@@ -123,8 +123,11 @@ class DownstreamSync(SyncProcess):
 
     @SyncProcess.error.setter
     def error(self, value):
-        if value is not None and self.pr:
-            env.gh_wpt.add_labels(self.pr, "mozilla:gecko-blocked")
+        if self.pr:
+            if value is not None:
+                env.gh_wpt.add_labels(self.pr, "mozilla:gecko-blocked")
+            else:
+                env.gh_wpt.remove_label(self.pr, "mozilla:gecko-blocked")
         return SyncProcess.error.fset(self, value)
 
     @property
--- a/sync/downstream.py
+++ b/sync/downstream.py
@@ -127,7 +127,7 @@ class DownstreamSync(SyncProcess):
             if value is not None:
                 env.gh_wpt.add_labels(self.pr, "mozilla:gecko-blocked")
             else:
-                env.gh_wpt.remove_label(self.pr, "mozilla:gecko-blocked")
+                env.gh_wpt.remove_labels(self.pr, "mozilla:gecko-blocked")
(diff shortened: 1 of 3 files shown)


sync/downstream.py
```diff
@@ -121,6 +121,12 @@ def pr_head(self):
         merged unless it happens to be a fast-forward"""
         return sync_commit.WptCommit(self.git_wpt, "origin/pr/%s" % self.pr)
 
+    @SyncProcess.error.setter
+    def error(self, value):
+        if value is not None and self.pr:
+            env.gh_wpt.add_labels(self.pr, "mozilla:gecko-blocked")
```
jgraham: We should also remove the label if its set and the value is `None`.

dhdavvie: The only problem I see there is in the case [here](https://github.com/mozilla/wpt-sync/blob/b15eb870b34f4e4e8d103f55d911c8046de68485/sync/downstream.py#L605) where `self.error = exception` and exception can be None. That doesn't mean that downstream has stopped failing if I understand it correctly.

We should probably have some sort of check that when all the try statuses are passing we can remove the label

sync/gh.py
```diff
@@ -82,6 +82,16 @@ def add_labels(self, pr_id, *labels):
         issue = self.repo.get_issue(pr_id)
         issue.add_to_labels(*labels)
 
+    def remove_labels(self, pr_id, *labels):
+        issue = self.repo.get_issue(pr_id)
+        for label in labels:
+            self.remove_label(pr_id, label, issue=issue)
+
+    def remove_label(self, pr_id, label, issue=None):
```
jgraham: I'm quite tempted to just inline this into `remove_label`, since `remove_labels` handles the case of a single label with just the same API.

dhdavvie: I've inline the methods into a single `remove_labels`

sync/gh.py
```diff
@@ -348,6 +354,10 @@ def set_status(self, pr_id, status, target_url, description, context):
     def add_labels(self, pr_id, *labels):
         self.get_pull(pr_id)["labels"].extend(labels)
 
+    def remove_labels(self, pr_id, *labels):
+        pr = self.get_pull(pr_id)
+        pr["labels"] = filter(lambda l: l not in labels, pr["labels"])
```
jgraham:
```suggestion
        pr["labels"] = [item for item in pr["labels"] if item not in labels]
```

PR #423, opened 2019-08-08
Handle ongoing landing better in wptsync landing|retrigger

jgraham:
If there's an ongoing landing, start looking for new landable commits after the head that
landing will update to

### DIFF
--- a/sync/command.py
+++ b/sync/command.py
@@ -510,23 +510,30 @@ def do_landable(git_gecko, git_wpt, *args, **kwargs):
     import update
     from sync import LandableStatus
     from downstream import DownstreamAction, DownstreamSync
-    from landing import load_sync_point, landable_commits, unlanded_with_type
+    from landing import current, load_sync_point, landable_commits, unlanded_with_type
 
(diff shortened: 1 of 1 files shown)


sync/command.py
```diff
@@ -510,23 +510,33 @@ def do_landable(git_gecko, git_wpt, *args, **kwargs):
...
-    from landing import load_sync_point, landable_commits, unlanded_with_type
+    from landing import current, load_sync_point, landable_commits, unlanded_with_type
 
+    current_landing = current(git_gecko, git_wpt)
     if kwargs["prev_wpt_head"] is None:
         sync_point = load_sync_point(git_gecko, git_wpt)
-        prev_wpt_head = sync_point["upstream"]
         print("Last sync was to commit %s" % sync_point["upstream"])
+
+    if current_landing:
```
dhdavvie: Wouldn't it make more sense to nest this if/else, and revert `if kwargs["prev_wpt_head"] is not None:` to an else? Looks like an extra branch of execution that could be completely irrelevant if `prev_wpt_head` is set

jgraham:
Updated.

Issue #394, opened 2019-07-22
docker build sometimes errors with `gpg: keyserver receive failed: Cannot assign requested address`


jgraham:
```
+ sudo apt-key adv --keyserver hkps.pool.sks-keyservers.net --recv-keys 0x6B73A36E6026DFCA
Warning: apt-key output should not be parsed (stdout is not a terminal)
Executing: /tmp/apt-key-gpghome.1VXI2TRoJH/gpg.1.sh --keyserver hkps.pool.sks-keyservers.net --recv-keys 0x6B73A36E6026DFCA
gpg: keyserver receive failed: Cannot assign requested address
```

PR #426, opened 2019-08-08
Retry apt-key command on travis if it fails

dhdavvie:
#394

We seem to have intermittent failure when running `sudo apt-key adv ...`. One way to mitigate this and hopefully less often have to manually restart builds is to retry the command if it fails, and if it  fails a certain number of times then we fail the build. 

[Here](https://travis-ci.org/dhdavvie/wpt-sync/jobs/569397682#L1708-L1735) is an example where the command failed twice, but finally on the third attempt it passed.

### DIFF
--- a/docker/Dockerfile.dev
+++ b/docker/Dockerfile.dev
@@ -34,15 +34,15 @@ RUN set -eux; \
 
 RUN set -eux; \
     count=0; \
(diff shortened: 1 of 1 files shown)


jgraham:
Need to rebase onto master

PR #429, opened 2019-08-09
Added check to ensure pr_id is converted to int

dhdavvie:
https://paste.mozilla.org/LCBhptWp

Also ensure proper handling in situations where we try to remove a label from an issue that doesn't have it.

### DIFF
--- a/sync/gh.py
+++ b/sync/gh.py
@@ -88,7 +88,13 @@ class GitHub(object):
         pr_id = self._convert_pr_id(pr_id)
         issue = self.repo.get_issue(pr_id)
         for label in labels:
-            issue.remove_from_labels(label)
+            try:
+                issue.remove_from_labels(label)
+            except github.GithubException as e:
+                if e.data["message"] != "Label does not exist":
+                    raise e
+                else:
+                    logger.warning("Tried to remove label %s that doesn't exist" % label)
 
     def _convert_pr_id(self, pr_id):
         if not isinstance(pr_id, (int, long)):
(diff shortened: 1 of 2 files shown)


sync/gh.py
```diff
@@ -79,15 +79,25 @@ def set_status(self, pr_id, status, target_url, description, context):
...
+        pr_id = self._convert_pr_id(pr_id)
         issue = self.repo.get_issue(pr_id)
         issue.add_to_labels(*labels)
 
     def remove_labels(self, pr_id, *labels):
         logger.debug("Removing labels %s from PR %s" % (labels, pr_id))
+        pr_id = self._convert_pr_id(pr_id)
         issue = self.repo.get_issue(pr_id)
         for label in labels:
             issue.remove_from_labels(label)
```
jgraham: We also need to handle the GithubException where the label already exists (and might be worth special casing the caller to not call this at all if the error is already `None`.

dhdavvie: Agreed, I've updated

sync/gh.py
```diff
@@ -88,7 +88,13 @@ def remove_labels(self, pr_id, *labels):
         pr_id = self._convert_pr_id(pr_id)
         issue = self.repo.get_issue(pr_id)
         for label in labels:
-            issue.remove_from_labels(label)
+            try:
+                issue.remove_from_labels(label)
+            except github.GithubException as e:
+                if e.data["message"] != "Label does not exist":
```
jgraham: I think we should just ignore the error if the label doesn't exist and log a warning in other cases (for bonus points report to new relic). I don't think we want to block the process on the failure to remove a label.

PR #436, opened 2019-08-13
Avoid adding duplicate comments and needinfos in some cases

jgraham:
When adding a needinfo check the flags for an existing similar-looking
needinfo request, and when adding a comment using the BugContext
frontend, perform a similar check that we aren't posting duplicate content.


sync/bug.py
```diff
@@ -286,3 +308,9 @@ def __setitem__(self, name, value):
     def needinfo(self, *requestees):
         for requestee in requestees:
             self.changes.append("Setting bug %s needinfo %s" % (self.bug_id, requestee))
+
+    def add_comment(self, text, check_dupe=True):
+        self.comment = text
```
mjzffr: I'm curious, if you include the "one comment per bug" ValueError here, do the tests still pass?

PR #438, opened 2019-08-14
Update test types run on try

jgraham:
pgo ccov and msvc builds no longer exist, but shippable builds are slow and
can be excluded from the PR runs.


sync/landing.py
```diff
@@ -894,7 +894,7 @@ def update_landing(git_gecko, git_wpt, prev_wpt_head=None, new_wpt_head=None,
                                        landing,
                                        hacks=False,
                                        try_cls=trypush.TryFuzzyCommit,
-                                       exclude=["pgo", "ccov", "msvc"])
+                                       exclude=[])
```
mjzffr: Also line 910?

PR #439, opened 2019-08-14
Check if the bot has merged a PR and comment on BMO

dhdavvie:
This functionality seems to be missing. By checking the 'merged_by' user we can see if the bot has merged the PR, and report back to BMO. Extended a test to also check this comment gets posted.

### DIFF
--- a/sync/handlers.py
+++ b/sync/handlers.py
@@ -55,6 +55,7 @@ def handle_pr(git_gecko, git_wpt, event):
 
         merge_sha = (event["pull_request"]["merge_commit_sha"]
                      if event["pull_request"]["merged"] else None)
+        merged_by = (event["pull_request"]["merged_by"]["login"] if merge_sha else None)
         with SyncLock.for_process(sync.process_name) as lock:
             with sync.as_mut(lock):
                 update_func(git_gecko,
@@ -62,7 +63,8 @@ def handle_pr(git_gecko, git_wpt, event):
                             sync,
                             event["action"],
                             merge_sha,
-                            event["pull_request"]["base"]["sha"])
+                            event["pull_request"]["base"]["sha"],
+                            merged_by)
 
 
 def handle_status(git_gecko, git_wpt, event):
--- a/sync/upstream.py
+++ b/sync/upstream.py
@@ -950,7 +950,7 @@ def commit_status_changed(git_gecko, git_wpt, sync, context, status, url, sha):
 
 @entry_point("upstream")
 @mut('sync')
-def update_pr(git_gecko, git_wpt, sync, action, merge_sha=None, base_sha=None):
+def update_pr(git_gecko, git_wpt, sync, action, merge_sha=None, base_sha=None, merged_by=None):
     """Update the sync status for a PR event on github
 
     :param action string: Either a PR action or a PR status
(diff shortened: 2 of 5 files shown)


sync/upstream.py
```diff
@@ -967,5 +967,8 @@ def update_pr(git_gecko, git_wpt, sync, action, merge_sha=None, base_sha=None):
             if sync.status not in ("complete", "wpt-merged"):
                 env.bz.comment(sync.bug, "Upstream PR merged")
                 sync.finish("wpt-merged")
+            elif (sync.status == "wpt-merged" and
+                    merged_by == env.config["web-platform-tests"]["github"]["user"]):
```
jgraham: Incorrect indent.

sync/handlers.py
```diff
@@ -55,14 +55,16 @@ def handle_pr(git_gecko, git_wpt, event):
...
         with SyncLock.for_process(sync.process_name) as lock:
             with sync.as_mut(lock):
                 update_func(git_gecko,
                             git_wpt,
                             sync,
                             event["action"],
                             merge_sha,
-                            event["pull_request"]["base"]["sha"])
+                            event["pull_request"]["base"]["sha"],
+                            merged_by)
```
jgraham: This can end up calling `downstream.update_pr` with an extra argument.

sync/upstream.py
```diff
@@ -967,5 +967,8 @@ def update_pr(git_gecko, git_wpt, sync, action, merge_sha=None, base_sha=None):
             if sync.status not in ("complete", "wpt-merged"):
                 env.bz.comment(sync.bug, "Upstream PR merged")
```
jgraham: Can we add the username here?

sync/upstream.py
```diff
@@ -967,5 +967,8 @@ def update_pr(git_gecko, git_wpt, sync, action, merge_sha=None, base_sha=None):
...
                 sync.finish("wpt-merged")
+            elif (sync.status == "wpt-merged" and
```
jgraham: Why don't we just do this when we call `merge_pr` rather than waiting for the event?

sync/upstream.py
```diff
@@ -967,5 +967,8 @@ def update_pr(git_gecko, git_wpt, sync, action, merge_sha=None, base_sha=None):
...
+                    merged_by == env.config["web-platform-tests"]["github"]["user"]):
+                env.bz.comment(sync.bug, "Upstream PR merged by me")
```
jgraham: Can we just use the username rather than `me`? I think it's clearer.

sync/upstream.py
dhdavvie: The linter seemed to complain about a shallower indent, will try again.

sync/handlers.py
dhdavvie: I have added a kwarg in the downstream.update_pr [here](https://github.com/mozilla/wpt-sync/pull/439/files#diff-9145f0a692affb6ee2e22ccd4014b221R1050) to avoid this.

sync/upstream.py
dhdavvie: Good idea

sync/upstream.py
dhdavvie: I did consider this, will move it over to that part of the logic.

PR #448, opened 2019-08-20
Handle bmo errors

dhdavvie:
Relic logs shows a high error rate at the moment to do with getting 0 length responses from BMO. Doing a broad except and raising retryable error. Maybe should be catching ValueError specifically for this?

### DIFF
--- a/sync/bug.py
+++ b/sync/bug.py
@@ -5,9 +5,9 @@ import traceback
 import urlparse
 
 import bugsy
+import newrelic
 
 from env import Environment
-from errors import RetryableError
 
 env = Environment()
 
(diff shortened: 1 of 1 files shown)


sync/bug.py
```diff
@@ -93,6 +94,10 @@ def _get_bug(self, bug_id):
             except bugsy.BugsyException:
                 logger.error("Failed to retrieve bug with id %s" % bug_id)
                 return
+            except Exception as e:
```
jgraham: I'm not sure what to do in this case; I see we already return None in some cases, but I kind of doubt that all the callers can really handle it. But maybe we should do that for now, plus log to NR, and fix this better when we are doing all the bug updates in a single place.

PR #458, opened 2019-08-26
Ensure we do a full metadata update in landings

jgraham:


jgraham:
@mjzffr Oh were it that simple! Landings end up retriggering any jobs that still fail after a metadata update. But more importantly try pushes are run with `--update-intermittent` and @nikk15 has ensured that running with `--full` doesn't destroy information about intermittents where the new status is compatible with the existing set of intermittents.

PR #462, opened 2019-08-27
[WIP] Get metadata from upstream Taskcluster run

dhdavvie:
Upstream PRs have a Taskcluster run which we could use as a temporary metadata source, with eventual goal being to reduce the amount of CI we run on our end. 

This is a first draft at how we could fetch the upstream wpt_report.json and store it locally. Next step would be looking into how we can make this replace the initial Try push. 

Notes:
- The report artifact is called `wpt_report.json` on Taskcluster, may need to rename it to be in line with the naming `wptreport.json` that we get from try?
- Will need to implement a way to cleanup so we do not have a space leak


sync/downstream.py
```diff
@@ -1069,3 +1075,32 @@ def update_pr(git_gecko, git_wpt, sync, action, merge_sha, base_sha, merged_by=N
     except Exception as e:
         sync.error = e
         raise
+
+
+def get_temporary_metadata(sync):
```
jgraham: `temporary` seems like a bad name here; the point is that it's from GitHub.

I also wonder if it would be more like the other code to make this a method on the sync.

sync/downstream.py
```diff
@@ -1069,3 +1075,32 @@ def update_pr(git_gecko, git_wpt, sync, action, merge_sha, base_sha, merged_by=N
...
+    pr_id = sync.pr
+    _, statuses = env.gh_wpt.get_combined_status(pr_id)
+    for status in statuses:
+        if status.context == "Taskcluster (pull_request)":
```
jgraham: Hmm, so this name might change if we ever switch on the checks API (which we previously tried to do and failed because wpt.fyi wasn't ready). I wonder if we can look for things with a taskcluster URL (although that will also change; we could/should make it a config variable).

sync/downstream.py
```diff
@@ -945,6 +946,10 @@ def commit_status_changed(git_gecko, git_wpt, sync, context, status, url, head_s
             sync.update_commits()
             sync.update_github_check()
             return
+
+        if context == "Taskcluster (pull_request)" and status == "success":
```
jgraham: (Comment about using the name only applies here)

sync/downstream.py
```diff
@@ -945,6 +946,10 @@ def commit_status_changed(git_gecko, git_wpt, sync, context, status, url, head_s
...
+            get_temporary_metadata(sync)
```
jgraham: We probably want to ensure we only do this once per push per sync.

sync/downstream.py
```diff
@@ -1069,3 +1075,32 @@ def update_pr(git_gecko, git_wpt, sync, action, merge_sha, base_sha, merged_by=N
...
+            if len(tasks) != 1:
+                logger.error("Could not find the TaskCluster task for %s" % task_name)
+                return False
+            get_wpt_report(tasks, pr_id)
+            try:
+                log_path = tasks.tasks[0]['status']['runs'][0]['_log_paths']['wpt_report.json']
+            except KeyError:
+                logger.warning("Log path not found for downloaded logs from PR Taskcluster run")
+                return False
+            sync.update_metadata([log_path])
```
jgraham: Some thought is required here; for new tests it's probably fine. For other cases we might want to worry about how platform differences are going to be handled. We should at least run the experiment by (by hand) taking a try push for a PR from the sync bot and updating it using the results from upstream and ensuring it's doing what we expect.

sync/tc.py
```diff
@@ -443,3 +444,38 @@ def cleanup():
                 now - timedelta(days=5)):
                 logger.info("Removing downloaded logs without recent activity %s" % rev_path)
                 shutil.rmtree(rev_path)
+
+
+def get_wpt_report(tasks, pr_id, session=None):
```
jgraham: Historically this kind of code hasn't been in tc.py itself (e.g. the existing code is in the try class). Could be a chance to refactor somewhat, so that we can share codepaths for downloading logs with the try case.

sync/tc.py
```diff
@@ -443,3 +444,38 @@ def cleanup():
...
+    task = tasks.tasks[0]
+    task_id = task.get('status', {}).get('taskId')
+    if not task_id:
+        logger.error("TaskCluster task not found for PR %s" % pr_id)
+        return
+    log_path = os.path.join(base_path, task_id + "_" + filename)
+    archive_path = os.path.join(base_path, task_id + "_" + archive)
+
+    # TODO maybe need to convert wpt_report to wptreport to be consistent with Try push versions?
+    with gzip.open(archive_path, 'rb') as f:
```
jgraham: An alternate option would be to make the metadata-update code handle gzipped input (by making a patch in mozilla-central). I think I prefer that idea if possible?

sync/downstream.py
dhdavvie: Doesn't this already check for this, as it is checking if the taskcluster run has completed and then attempts to get the metadata.

sync/downstream.py
dhdavvie: Should we disable this line for the time being and so that we can check manually till we are confident?

sync/tc.py
dhdavvie: I have been toying with the idea of making a TryPush like class for these Taskcluster runs, will look into how this could work

sync/downstream.py
jgraham: That makes sense.

sync/downstream.py
dhdavvie: I've added a config under `[web-platform-tests]` `ci.context` in the `sync.ini` files

sync/trypush.py
```diff
@@ -450,6 +451,71 @@ def delete(self):
                 idx.delete(key, data)
 
 
+class TcPush(TryPush):
```
jgraham: So overall, having seen it, this feels a little weird. The other classes that we store as top-level objects are things that where we're (at least conceptually) in control of the lifecycle. But in this case we aren't in control and just have a single point of interaction with the taskcluster run, so the only thing that we do with this object is call `download_logs` and it's otherwise exposing a bunch on unused API surface, at least some of which doens't make sense.

I think we should just keep the `download_logs` function, but otherwise remove this class. If we want to store metadata about which taskgroups we've previously processes, it might make sense to have `upstream_tc_taskgroup` on the sync object, for example.

jgraham: So, now having actually looked at the code, I'm not convinced that this is the right approach.

We typically only store these `ProcessData`-derived objects in cases where we are actually managing the lifecycle of some process. In this case we aren't but just intereacting with it at a point. That means a lof of the API here is unused; in practice we only call `download_logs`.

Unless we are planning to use a lot more of this API surface in the future, it seems like we should just factor out the `download_logs` method into a function and, if we want to store the taskgroup_id, put it on the sync object directly.

dhdavvie: I agree, it does feel out of place in this use case. Going to remove this change, and have a function in `trypush.py` for downloading the logs.

sync/downstream.py
```diff
@@ -1057,6 +1103,7 @@ def update_pr(git_gecko, git_wpt, sync, action, merge_sha, base_sha, merged_by=N
         elif action == "closed":
             # We are storing the wpt base as a reference
             sync.data["wpt-base"] = base_sha
+            sync.get_gh_metadata()
```
jgraham: I"m not sure why it's called here?

sync/downstream.py
```diff
@@ -811,6 +812,47 @@ def update_metadata(self, log_files, stability=False):
...
+
+                try:
+                    # TODO replace lines when update_metadata enabled
+                    # log_path =tasks.tasks[0]['status']['runs'][0]['_log_paths']['wpt_report.json']
+                    tasks.tasks[0]['status']['runs'][0]['_log_paths']['wpt_report.json']
+                except KeyError:
+                    logger.warning("Log path not found for downloaded logs from PR Taskcluster run")
+                    return False
+                # TODO Disabled till we are sure it is working correctly
+                # self.update_metadata([log_path])
```
jgraham: I think in any case this call should be outside this code.

sync/downstream.py
```diff
@@ -811,6 +812,47 @@ def update_metadata(self, log_files, stability=False):
 
         return disabled
 
+    @mut()
+    def get_gh_metadata(self):
```
jgraham: I think this should be `download_upstream_logs` or something that doesn't contain the string `metadata` since that's really the result of updating from the logs, not the logs themselves.

sync/downstream.py
dhdavvie: I think for the time being this makes sense here whilst we are trialing this, in the future it would make more sense for it to be in some refactored `try_notify` that also handles the notification of upstream CI tasks. 

This does result in a space leak currently, since we do not process the logs and remove them.

sync/downstream.py
dhdavvie: Agreed, I think it will be better once the try and upstream CI process is merged. Will remove for now and we can figure out where it will go in the future

dhdavvie:
Closing this for now as it looks like this is not the right direction.

PR #472, opened 2019-09-04
lower rebuild count, use config to set rebuild count

dhdavvie:


sync/trypush.py
```diff
@@ -191,7 +191,7 @@ def create(cls, lock, sync, affected_tests=None, stability=False, hacks=True,
         git_work = sync.gecko_worktree.get()
 
         if rebuild_count is None:
-            rebuild_count = 0 if not stability else 10
+            rebuild_count = 0 if not stability else env.config['gecko']['try']['stability_count']
```
jgraham: Can we give this a default value if the configuration isn't found (and log an error or something)?

dhdavvie: good idea

Issue #466, opened 2019-08-29
Seeing a lot of unexplained mach failures


jgraham:
This is happening trying to update metadata.

c.f. https://rpm.newrelic.com/accounts/1872822/applications/43012201/filterable_errors#/show/c168b5ab-ca57-11e9-8523-0242ac110008_0_6599/stack_trace?top_facet=transactionUiName&primary_facet=error.class&barchart=barchart

It would be good to figure out root causes here as this is a thing that blocks sync from being landing-ready. The first step would be to improve the logging so that the underlying mach error ends up being recorded rather than (or in addition to) the resulting stack in the sync code.

dhdavvie:
https://rpm.newrelic.com/accounts/1872822/applications/43012201/traced_errors/81ff6065-d9fb-11e9-b4ad-0242ac110007_0_6932

Unfortunately this didn't seem to help, as the output gets truncated by NR. Could try printing the tail end of the output, but again this may just cut out useful information. Do we have an alternative place to log this? We could log it and manually check the logs on the server but that seems like a painful process.

jgraham:
OK, so running things interactively it seems like we sometimes end up with corrupted caches in the `~/.mozbuild/` directory. That seems like it could be a wpt issue but maybe we should delete that directory in the cleanup, or when we get an error.

dhdavvie:
This is addressed in #500.

PR #480, opened 2019-09-09
Record CalledProcessErrors better with newrelic, continue raising.

dhdavvie:
#466

The output of recorded Newrelic log seems to be truncated due to length. This should hopefully provide us with a field to see the relevant information about the error.

### DIFF
--- a/sync/projectutil.py
+++ b/sync/projectutil.py
@@ -13,6 +13,8 @@ import types
 
 import newrelic
 
+from .errors import AbortError
+
 logger = logging.getLogger(__name__)
 
 
@@ -39,8 +41,10 @@ class Command(object):
         try:
             return subprocess.check_output(command, cwd=self.path, **opts)
         except subprocess.CalledProcessError as e:
-            newrelic.agent.record_exception(params={"mach_output": e.output})
-            raise e
+            newrelic.agent.record_exception(params={
+                "command": self.name,
+                "command_output": e.output})
+            raise AbortError("Problem running '%s' command" % self.name)
 
     def __getattr__(self, name):
         if name.endswith("_"):
--- a/sync/projectutil.py
+++ b/sync/projectutil.py
@@ -13,8 +13,6 @@ import types
 
 import newrelic
 
-from .errors import AbortError
-
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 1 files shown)


sync/projectutil.py
```diff
@@ -34,7 +36,11 @@ def get(self, *subcommand, **opts):
         assert subcommand and len(subcommand)
         command = [os.path.join(self.path, self.name)] + list(subcommand)
         logger.info("Running command:\n %s" % " ".join(command))
-        return subprocess.check_output(command, cwd=self.path, **opts)
+        try:
+            return subprocess.check_output(command, cwd=self.path, **opts)
+        except subprocess.CalledProcessError as e:
+            newrelic.agent.record_exception(params={"mach_output": e.output})
```
jgraham: This isn't always `mach`. Also I'm worried that the exception is going to end up being recorded twice, unless it's always caught and replaced by an `AbortError`.

dhdavvie: Ahh yes, forgot I was changing the overall command class. I think replacing it with an AbortError may be the way to go, will adjust commit.

sync/projectutil.py
```diff
@@ -39,8 +41,10 @@ def get(self, *subcommand, **opts):
         try:
             return subprocess.check_output(command, cwd=self.path, **opts)
         except subprocess.CalledProcessError as e:
-            newrelic.agent.record_exception(params={"mach_output": e.output})
-            raise e
+            newrelic.agent.record_exception(params={
+                "command": self.name,
```
jgraham: Can we also log the exit code, just in case it's interesting?

sync/projectutil.py
```diff
@@ -39,8 +41,10 @@ def get(self, *subcommand, **opts):
...
+                "command_output": e.output})
+            raise AbortError("Problem running '%s' command" % self.name)
```
jgraham: I think on balance suggesting changing this to `AbortError` was wrong. There's actually no special handling for `AbortError` in the error reporting, and there's a place where we catch the `CalledProcessError` and have command-specific handling. 

In theory it would be better to have this handling for each instance of the command, as some cases might not be "exceptional" errors. But that's a lot of work and probably not worthwhile.

So I think reverting this back to just `raise` is actually the best trade-off. Sorry.

PR #481, opened 2019-09-09
Handle missing comment field in bug.

jgraham:
The error handling here caused some unnecessary exceptions; make it more
lenient in case we have code that sets and unsets the comment or something


dhdavvie:
LGTM, although in which cases do we set and then unset a comment? Could be a place to look at handling setting and unsetting properly in the future

jgraham:
Turns out this change was just nonsense. I made a better one instead.

PR #483, opened 2019-09-09
Use --update-intermittent unconditionally for landings

jgraham:


sync/landing.py
```diff
@@ -584,11 +584,7 @@ def update_metadata(self, log_files, update_intermittents=False):
         logger.info("Updating metadata from %s logs" % len(log_files))
         args = ["--full"]
         if update_intermittents:
-            help_text = mach.wpt_update("--help")
-            if "--update-intermittent " in help_text:
-                args.append("--update-intermittent")
-            # We didn't use --stability here since it's unclear which
-            # bug it will be associated with
+            args.append("--update-intermittent")
```
dhdavvie: Why is it that we can now remove this safety check?

jgraham: Because on mozilla-central we now know that the argument is supported; we can't possibly have a revision where it still isn't supported.

PR #484, opened 2019-09-09
Handle wpt tests-affected failing.

jgraham:
This can happen if we accidentially broke the command. Just assume that nothing is
affected in this case.

### DIFF
--- a/sync/downstream.py
+++ b/sync/downstream.py
@@ -8,6 +8,7 @@ from __future__ import print_function
 
 import os
 import re
+import subprocess
 import traceback
 from collections import defaultdict
 from datetime import datetime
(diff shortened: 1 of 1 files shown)


sync/downstream.py
```diff
@@ -758,7 +759,11 @@ def affected_tests(self, revish=None):
             if revish:
                 args.append(revish)
             logger.info("Getting a list of tests affected by changes.")
-            output = self.wpt.tests_affected(*args)
+            try:
+                output = self.wpt.tests_affected(*args)
+            except subprocess.CalledProcessError:
+                logger.error("Calling wpt tests-affected failed")
+                return tests_by_type
```
dhdavvie: Should we not be setting `self.data["affected-tests"]` like we did before? or are we ignoring this attempt all together if it fails?

sync/landing.py
```diff
@@ -961,6 +961,7 @@ def try_push_complete(git_gecko, git_wpt, try_push, sync, allow_push=True,
     intermittents = []
 
     if not accept_failures and try_push.status == "complete":
+        sync.next_try_push()
```
dhdavvie: Why are we doing this here?

dhdavvie: Looks like it was supposed to be part of #482 maybe?

sync/downstream.py
jgraham: I think it's OK to just ignore it if it fails.

PR #500, opened 2019-09-24
Remove .mozbuild wpt cache

jgraham:
This prevents the cache becoming a space leak, and avoids a somewhat
common error we've seen with the cache becoming corrupted. The reasons
for this are unclear, but the performance implications of simply not
using the cache are well within the bounds of what's reasonable to get
a simple fix.


dhdavvie:
Looks like build is failing due to dependency updates: https://pypi.org/project/configparser/#history

Will wait for the dependency update PR

PR #512, opened 2019-10-03
Remove initial try push & push stability once PR is merged.

dhdavvie:
With the current implementation, the bot can spam the try server with try pushes from an upstream PR (see [here](https://bugzilla.mozilla.org/show_bug.cgi?id=1573920)). In some cases with the Chromium export bot, it also pushes stability try runs because the upstream PR is already approved but still receiving changes from the downstream Chromium repo. 

By removing the initial try push entirely and only pushing a stability run once the PR has been merged, we can reduce a significant amount of spam and save on cost. The trade-off is turn around time of the sync, but this should not be that significant.

Here I've changed `next_action` to go straight to stability run on pr merged, and the same has been done in `LandingSync.next_try_push`. Log downloading has been changed to handle stability runs by adding the names of tasks to the exclude list, which should in theory prevent the downloading of logs from repeat runs. Tests have also been updated to reflect this new workflow.

### DIFF
--- a/sync/downstream.py
+++ b/sync/downstream.py
@@ -819,7 +819,7 @@ class DownstreamSync(SyncProcess):
 
         complete_try_push = None
         for try_push in sorted(self.try_pushes(), key=lambda x: -x.process_name.seq_id):
-            if try_push.status == "complete" and not try_push.stability:
+            if try_push.status == "complete":
                 complete_try_push = try_push
                 break
 
@@ -836,7 +836,8 @@ class DownstreamSync(SyncProcess):
         with try_push.as_mut(self._lock):
             try_tasks = complete_try_push.tasks()
             try:
-                complete_try_push.download_logs(try_tasks.wpt_tasks, report=True, raw=False)
+                complete_try_push.download_logs(try_tasks.wpt_tasks, report=True,
+                                                raw=False, first_only=True)
             except RetryableError:
                 logger.warning("Downloading logs failed")
                 return
--- a/sync/downstream.py
+++ b/sync/downstream.py
@@ -171,8 +171,9 @@ class DownstreamSync(SyncProcess):
         if latest_try_push.status != "complete":
             return DownstreamAction.wait_try
 
-        # Don't worry about recreating stability try pushes for infra failures
-        if latest_try_push.infra_fail and len(self.latest_busted_try_pushes()) > 5:
+        # If we have infra failure, flag for human intervention. Retrying stability runs would be
+        # very costly
+        if latest_try_push.infra_fail:
             return DownstreamAction.manual_fix
 
         return DownstreamAction.ready
--- a/sync/downstream.py
+++ b/sync/downstream.py
@@ -43,7 +43,7 @@ class DownstreamAction(enum.Enum):
     try_push = 2
     try_push_stability = 3
     wait_try = 4
-    wait_approved = 5
+    wait_upstream = 5
 
     def reason_str(self):
         return {DownstreamAction.ready: "",
@@ -51,7 +51,7 @@ class DownstreamAction(enum.Enum):
                 DownstreamAction.try_push: "valid try push required",
                 DownstreamAction.try_push_stability: "stability try push required",
                 DownstreamAction.wait_try: "waiting for try to complete",
-                DownstreamAction.wait_approved: "waiting for PR to be approved"}.get(self, "")
+                DownstreamAction.wait_upstream: "waiting for PR to be merged"}.get(self, "")
 
 
 class DownstreamSync(SyncProcess):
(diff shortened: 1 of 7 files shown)


sync/downstream.py
```diff
@@ -159,30 +159,22 @@ def next_action(self):
...
-            logger.debug("Sync for PR %s has no valid try push" % self.pr)
-            return DownstreamAction.try_push
+            if self.requires_stability_try:
+                logger.debug("Sync for PR %s requires a stability try push" % self.pr)
+                pr = env.gh_wpt.get_pull(self.pr)
+                if pr.merged:  # Wait till PR is merged to push stability run
+                    return DownstreamAction.try_push_stability
+                return DownstreamAction.wait_approved
+            else:
+                return DownstreamAction.try_push
```
dhdavvie: This I am a bit unsure what to do here. I left it as a `try_push` state because this is now obsolete anyway and won't do anything. Would the ready state be more appropriate? Since if it does not require stability then there is nothing else to be done

sync/downstream.py
```diff
@@ -277,14 +269,7 @@ def next_try_push(self, try_cls=trypush.TryFuzzyCommit):
         self.affected_tests()
 
         action = self.next_action
-        if action == DownstreamAction.try_push:
```
jgraham: I think we can keep this? Something like an infra change doesn't require a stability try push, but we should push to try to ensure the world didn't break

sync/downstream.py
```diff
@@ -159,30 +159,22 @@ def next_action(self):
         latest_try_push = self.latest_valid_try_push
 
         if not latest_try_push:
-            logger.debug("Sync for PR %s has no valid try push" % self.pr)
-            return DownstreamAction.try_push
+            if self.requires_stability_try:
+                logger.debug("Sync for PR %s requires a stability try push" % self.pr)
+                pr = env.gh_wpt.get_pull(self.pr)
+                if pr.merged:  # Wait till PR is merged to push stability run
```
jgraham: I think this condition should be outside the initial `if` i.e. if we're not merged always return `wait_approved` (which could be renamed `wait_upstream` or something now it's not about being approved).

sync/downstream.py
jgraham: No, I think only doing a simple try push where there are no test changes but are infra changes still makes sense.

sync/trypush.py
```diff
@@ -400,7 +400,14 @@ def included(t):
             # if a name is on the excluded list, only download SUCCESS job logs
             name = t.get("task", {}).get("metadata", {}).get("name")
             state = t.get("status", {}).get("state")
-            return name not in exclude or state == tc.SUCCESS
+            output = name not in exclude or state == tc.SUCCESS
+
+            # Add this name to exclusion list, so that we don't download the repeated run
+            # logs in a stability try run
+            if stability and name is not None and name not in exclude:
```
jgraham: I don't understand why you are excluding the repeated run logs? We need those if we want to update the metadata to record intermittent tests correctly.

sync/downstream.py
```diff
@@ -1042,6 +1028,7 @@ def try_push_complete(git_gecko, git_wpt, try_push, sync):
 @entry_point("downstream")
 @mut('sync')
 def pull_request_approved(git_gecko, git_wpt, sync):
+    # We no longer run Try pushes when a GH PR has been approved, so this is sort of useless now.
```
jgraham: OK, so let's just remove this method and callers.

sync/trypush.py
dhdavvie: So this was mainly to avoid recording all the logs for when we use `try_notify`. Otherwise the bot will notify with a count of all the tests run including any repeats, which may not be the most useful notification? This is off by default for all other pieces of functionality in wpt-sync.

sync/downstream.py
dhdavvie: That sounds like a good idea, will also need to move the `manual_fix` and `wait_try` states into this new if.

sync/downstream.py
dhdavvie: Yes that is true, will reimplement the normal try push for this situation

sync/trypush.py
jgraham: In that case we should probably call it `first_only` or `one_per_job` or something. In practice I think the notification code only works with one log per job name so at most we end up redoing some work. Still it seems like it would be better if we actually tried to use the data to tell people about any instabilities, but agreed that we shouldn't block this change on that.

dhdavvie: Renaming it sounds like a good idea. I think we should keep this for now, and file an issue about improving the notification system to provide better metrics using these stability runs.

sync/gh.py
```diff
@@ -374,6 +374,12 @@ def remove_labels(self, pr_id, *labels):
         pr = self.get_pull(pr_id)
         pr["labels"] = [item for item in pr["labels"] if item not in labels]
 
+    def load_pull(self, data):
+        pr = self.get_pull(data["number"])
+        pr.merged = data["merged"]
+        pr.state = data["state"]
+        pass
```
jgraham: Hmm?

sync/landing.py
```diff
@@ -617,20 +617,14 @@ def next_try_push(self, retry=False):
             return
 
         latest_try_push = self.latest_try_push
-        stability = False
+        stability = True
```
jgraham: I don't think we want to change the landing code here. We still need to do two try pushes there.

dhdavvie: Ahh I see, I was under the impression we also wanted that changed. Will revert!

PR #515, opened 2019-10-03
Add results notifications based on wpt.fyi data

jgraham:
This is a simple implementation of wpt.fyi-based notifications in addition to those coming from Gecko data. So far we treat the two data sources entirely seperately; we may revisit that for clarity later on.


jgraham:
This is missing tests; I'll add those.

sync/notify/__init__.py
```diff
@@ -0,0 +1,44 @@
...
+def truncate_message(parts):
+    # This is currently the number of codepoints, with the proviso that
+    # only BMP characters are supported
+    MAX_LENGTH = 65535
+
+    suffix = "\n(truncated for maximum comment length)"
+    message = ""
+    for part in parts:
+        if len(message) + len(part) + 1 > MAX_LENGTH - len(suffix):
+            message += suffix
```
dhdavvie: If the first part is over the max length, then the message will only contain the truncated notification and nothing else. Should we not max out the message with content if possible?

sync/downstream.py
```diff
@@ -832,36 +832,29 @@ def try_notify(self):
...
-                complete_try_push.download_logs(try_tasks.wpt_tasks, report=True, raw=False)
-            except RetryableError:
-                logger.warning("Downloading logs failed")
-                return
-
-        msg = notify.get_msg(try_tasks.wpt_tasks, central_tasks)
-        env.bz.comment(self.bug, msg)
+        with env.bz.get_ctx(self.bug) as bug:
+            if truncated:
+                # TODO: Upload the full message as an attachment
```
dhdavvie: Is that not what this is doing?

sync/notify/__init__.py
jgraham: In practice that can't happen because the first part is never long enough. But I'll change the truncation text to mention the attachment so that it's useful even in that unlikley case.

sync/notify/__init__.py
```diff
@@ -0,0 +1,48 @@
...
+
+    suffix = "(See attachment for full changes)"
+    message = ""
+    truncated = False
+
+    padding = len(suffix) + 1
+    for part in parts:
+        if len(message) + len(part) + 1 > MAX_LENGTH - padding:
+            truncated = True
+            part = suffix
```
dhdavvie: Wouldn't it be better to do something like `part += '\n' + suffix` here? So that we maximise the amount of content in the comment

jgraham: Well no, because that would lead to an over-long comment. But we could do something like `part = part[:MAX_LENGTH - len(message) - padding] + '\n' + suffix`, but I'm not sure that having something cut off in the middle of a word is really better than just skipping the entire entry and hoping people look at the attachment.

dhdavvie: ahh yes sorry thats what I meant. I think either way is fine, I am just thinking that the more detail we can provide in the comment may lead to people not always having to download the attachment?

jgraham: Well hopefully having 65kB of text is pretty rare in the first place; that's a short novella worth of "important' failures to wade through.

jgraham: To sum up: you may be correct, but I don't think this should block landing this change. It's probably more useful to do followup work to condense the output than fiddle at the edges in terms of how we present too-long-for-a-comment data.

dhdavvie: Agreed!

PR #517, opened 2019-10-07
Use landable_status to check if sync is landable

jgraham:


sync/landing.py
```diff
@@ -817,7 +817,7 @@ def upstream_sync(bug_number):
                 logger.info("PR %s has no corresponding sync" % pr)
                 last = True
             elif (isinstance(sync, downstream.DownstreamSync) and
-                  not (sync.skip or sync.metadata_ready)):
+                  sync.landable_status != LandableStatus.ready):
                 logger.info("Metadata pending for PR %s" % pr)
```
dhdavvie: This log message should maybe also be updated to reflect that the sync could either be: skipped, metadata not ready or errored

jgraham: Fixed.

PR #521, opened 2019-10-08
Allow posting comments as Markdown

jgraham:


sync/bug.py
```diff
@@ -115,13 +115,17 @@ def _get_bug(self, bug_id):
...
-            logger.error("Failed to find bug %s to add comment:\n%s" % (bug_id, text))
+            logger.error("Failed to find bug %s to add comment:\n%s" % (bug_id, comment))
             return
-        text = check_valid_comment(text)
-        bug.add_comment(text)
+        body = {
+            "comment": check_valid_comment(comment)
+        }
+        body.update(kwargs)
+        self.bugzilla.bugzilla.request('bug/{}/comment'.format(bug.id),
```
dhdavvie: This new change in workflow means that anything calling `bz.comment` will instantly post the comment, instead of waiting for the context manager to exit. Is this something we want to do?

jgraham: We only call this in `__exit__` afaict?

dhdavvie: Searching the codebase finds 25 matches for `env.bz.comment`, so I don't think this is only called on `__exit__`

jgraham: Right, but those other uses were always synchronous; they just called `bug.add_comment` which is basically doing the same thing we do here but without the possibility to pass in other options.

dhdavvie: ahh I see what I was misunderstanding. I was confusing the `sync.bug.BugContext.add_comment` with `bugsy.bug.Bug.add_comment`. Now this makes a lot more sense! Yes this is an improvement on that functionality.

jgraham: We are slowly replacing bugsy with our own library with better functionality, but we aren't quite there yet ;)

PR #524, opened 2019-10-08
Enable passing multiple arbitary queries to mach try fuzzy

jgraham:
Use this to implement a fisson query that more closely matches the mochitest central
configuration.

### DIFF
--- a/sync/landing.py
+++ b/sync/landing.py
@@ -638,7 +638,9 @@ MANUAL PUSH: wpt sync bot
                                rebuild_count=0,
                                try_cls=trypush.TryFuzzyCommit,
                                full=True,
-                               exclude=["devedition", "ccov"])
+                               queries=["web-platform-tests !devedition !ccov !fis",
+                                        "web-platform-tests fis !devedition !ccov !asan "
+                                        "windows10 | linux64"])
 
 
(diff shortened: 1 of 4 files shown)


sync/landing.py
```diff
@@ -638,7 +638,9 @@ def next_try_push(self, retry=False):
                                rebuild_count=0,
                                try_cls=trypush.TryFuzzyCommit,
                                full=True,
-                               exclude=["devedition", "ccov"])
+                               queries=["web-platform-tests !devedition !ccov !fis",
+                                        "web-platform-tests fis !devedition !ccov !asan "
```
dhdavvie: Maybe I am misunderstanding Mach, but the help text for `mach try fuzzy` says for argument -q: `Specifying multiple times schedules the union of computed tasks.`. Won't this result in a decision task like `web-platform-tests !devedition !ccov !fis web-platform-tests fis !devedition !ccov !asan`, resulting in issues? notably having `!fis` and `fis`.

jgraham: No, it basically runs the queries one after another and adds in all the tasks selected by any query. You can try it locally in a gecko checkout with `./mach try fuzzy --no-push <queries>`.

PR #525, opened 2019-10-08
Weaker infra failure conditions, notify on build failures

dhdavvie:
If we have build failures, we should notify the bug. We should also not flag the try push for manual fix if we have some builds and subsequent tests that completed. 

Should also add build failure notification in `try_notify` but will probably do that in another PR.

### DIFF
--- a/sync/downstream.py
+++ b/sync/downstream.py
@@ -176,7 +176,13 @@ class DownstreamSync(SyncProcess):
             # If we have infra failure, flag for human intervention. Retrying stability
             # runs would be very costly
             if latest_try_push.infra_fail:
-                return DownstreamAction.manual_fix
+                tasks = latest_try_push.tasks()
+
+                # Check if we had any successful tests
+                if tasks.has_completed_tests():
+                    return DownstreamAction.ready
+                else:
+                    return DownstreamAction.manual_fix
 
             return DownstreamAction.ready
         else:
--- a/sync/trypush.py
+++ b/sync/trypush.py
@@ -205,6 +205,7 @@ class TryPush(base.ProcessData):
             "gecko-head": sync.gecko_commits.head.sha1,
             "wpt-head": sync.wpt_commits.head.sha1,
             "status": "open",
+            "bug": sync.bug,
         }
         process_name = base.ProcessName.with_seq_id(sync.git_gecko,
                                                     cls.obj_type,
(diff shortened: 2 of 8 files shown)


sync/trypush.py
```diff
@@ -350,19 +351,44 @@ def stability(self):
...
+        msg += "\n".join(task.get("task", {}).get("metadata", {}).get("name")
+                         for task in failed.tasks)
+        env.bz.comment(self.bug, msg)
+
     @infra_fail.setter
     @mut()
     def infra_fail(self, value):
-        """Is the current try push a stability test"""
+        """Set the status of this push's infrastructure failure state"""
         self["infra-fail"] = value
```
jgraham: Can we return if the value isn't changing?

sync/trypush.py
```diff
@@ -554,6 +585,14 @@ def has_failures(self):
             return any(task.get("status", {}).get("state") == tc.FAIL for task in wpt_tasks)
         return False
 
+    def has_completed_tests(self):
+        """Check if all the wpt tasks in a try push completed"""
+        wpt_tasks = self.wpt_tasks.filter(tc.is_test)
+        if wpt_tasks:                                          # tc.EXCEPTION ??
```
jgraham: Nah, if they all end up in Exception that's pretty bad. So this is fine.

sync/trypush.py
dhdavvie: Would be a good idea, will implement that.

sync/downstream.py
```diff
@@ -176,7 +176,13 @@ def next_action(self):
             # If we have infra failure, flag for human intervention. Retrying stability
             # runs would be very costly
             if latest_try_push.infra_fail:
-                return DownstreamAction.manual_fix
+                tasks = latest_try_push.tasks()
+
+                # Check if we had any successful tests
+                if tasks.has_completed_tests():
```
jgraham: I think this should use `tasks.wpt_tasks` to allow for the possibility of `source-test` jobs being considered tests.

dhdavvie: The `tasks.has_completed_tests` method filters for wpt tests like so `wpt_tasks = self.wpt_tasks.filter(tc.is_test)`. Will this filter out `source-test` jobs?

jgraham: Oh it seems like they're kind `source-test` so I guess this is fine.

PR #526, opened 2019-10-08
Ensure we don't create notification before try push is done

jgraham:
Before we were creating the notification after the non-stability try push.
But now we want to create it after the sync is ready to land

### DIFF
--- a/sync/downstream.py
+++ b/sync/downstream.py
@@ -1014,17 +1014,15 @@ def try_push_complete(git_gecko, git_wpt, try_push, sync):
                                               "\n".join(disabled)))
 
             try_push.status = "complete"
-            next_try_push = sync.next_try_push()
+            sync.next_try_push()
         finally:
             sync.update_github_check()
     else:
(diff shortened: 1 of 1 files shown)


sync/downstream.py
```diff
@@ -1040,7 +1038,6 @@ def update_pr(git_gecko, git_wpt, sync, action, merge_sha, base_sha, merged_by=N
             # We are storing the wpt base as a reference
             sync.data["wpt-base"] = base_sha
             sync.next_try_push()
-            sync.try_notify()
```
dhdavvie: Would it not be a good idea to keep this here? try_notify checks if results have already been notified so worse case it is a redundant call that returns immediately. But in case the try push completetion got missed, this would maybe be a useful fallback?

jgraham: No because now the PR is merged before we run try, so we were creating notifications without the try results being available.

dhdavvie: Yes of course, thanks

Issue #527, opened 2019-10-09
TryPush taskgroup_id is None


dhdavvie:
https://rpm.newrelic.com/accounts/1872822/applications/43012201/traced_errors/216b074b-e9d1-11e9-ab7c-0242ac11000c_0_4151

It seems as though when handling a try push complete, sometimes the `taskgroup_id` of the try push is None, despite being found by its `taskgroup_id` [here](https://github.com/mozilla/wpt-sync/blob/HEAD@%7B2019-10-09T10:00:58Z%7D/sync/handlers.py#L273). 

An initial fix would be to have the handler code check the taskgroup id of the try push and if it is None it simply sets it. 

We should also find out why try pushes are being saved by their taskgroup id successfully but never get that attribute set in them.

PR #528, opened 2019-10-09
Check a try push's taskgroup ID for None or errors

dhdavvie:
#527

### DIFF
--- a/sync/handlers.py
+++ b/sync/handlers.py
@@ -280,6 +280,25 @@ class TaskGroupHandler(Handler):
 
         with SyncLock.for_process(sync.process_name) as lock:
             with sync.as_mut(lock), try_push.as_mut(lock):
+                # We sometimes see the taskgroup ID being None. If it isn't set but found via its
+                # taskgroup ID, it is safe to set it here.
+                if try_push.taskgroup_id is None:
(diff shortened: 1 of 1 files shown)


sync/handlers.py
```diff
@@ -280,6 +280,18 @@ def __call__(self, git_gecko, git_wpt, body):
 
         with SyncLock.for_process(sync.process_name) as lock:
             with sync.as_mut(lock), try_push.as_mut(lock):
+                if try_push.taskgroup_id is None:
```
jgraham: Can you add a comment here explaining that this is actually a thing that happens? I wonder if we should also record an error in new relic in this case or something so we can track it more easily.

dhdavvie: Will add a comment. I was thinking we could use [record_custom_event](https://docs.newrelic.com/docs/agents/python-agent/python-agent-api/record_custom_event) but it doesn't look like that goes to the same general area that the Errors go to, instead going to the Insights tab? not sure how useful that is

jgraham: Hmm, I haven't really used the insights tab much, so I guess it's OK to skip that for now if you aren't moved to try it.

dhdavvie: I will look into it, it looks like it could be useful throughout the codebase

PR #536, opened 2019-10-14
Use Trusty Ubuntu distribution to speed up tests

dhdavvie:
It seems as though the change to the Xenial distribution of Ubuntu as
the default distribution for Travis made our tests run slower. This
should resolve that issue by locking the Ubuntu version to the Trust
version.

(Also more_itertools seems to have become a dependency due to locking the version)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+dist: trusty
 language: python
 python:
   - "2.7.14"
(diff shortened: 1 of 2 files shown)


jgraham:
It would be nice if we understood what was going on here, but if it reduces the test time by a factor of 2-3 that seems like something we should take right now.

dhdavvie:
I have not been able to find a concrete reason, but have seen some reports of 16.04 and 18.04 versions of Ubuntu causing slow downs in Docker containers. https://github.com/docker/for-linux/issues/738#issuecomment-523702651

qeternity:
AppArmor

PR #540, opened 2019-10-15
Add bug numbers to notifications based on wpt.fyi metadata

jgraham:
This adds read-only bug numbers to the notification where we think there's an existing bug covering the failure. This is a first step in more comprehensive support for creating bugs and keeping the metadata up to date with bugzilla-side changes.


sync/notify/wptfyimsg.py
```diff
@@ -156,7 +160,29 @@ def add_result(result, test_name, subtest_name):
...
+    specific_metadata = []
+    for item in test_metadata:
+        if "results" not in item:
+            specific_metadata.append(item)
+        else:
+            specific_results = [result for result in item["results"]
+                                if (("subtest" not in item or
+                                     item["subtest"] == subtest) and
+                                    ("status" not in item or
+                                     item["status"] == status))]
```
dhdavvie: This is very difficult to read, is there any way to make this easier to understand?

jgraham: Pushed something; idk if it's what you're looking for

dhdavvie: Looks much better to me, thanks!

PR #557, opened 2019-10-30
Add support for accessing wpt-metadata repo directly

jgraham:
wpt-metadata is a repo containing YAML files. Since we already have
good support for working with git repos, it makes sense to expediate
our ability to edit this data by supporting direct reads and writes
from the repository rather than needing to go via a wpt.fyi API.

The write support is designed to commit changes lazilly. This means
that we should be able to avoid text-based merge resolution by running
an algorithm where we pull the latest version of the upstream, apply
our edits, and then push. If the push fails due to conflicts we pull
again and retry.

So far this isn't wired up to any users in the repository; it will
eventually be used to support filing bugs for untriaged failures, and
for ensuring that the triage data is kept up to date with bugzilla
changes.

### DIFF
--- a/sync/meta.py
+++ b/sync/meta.py
@@ -1,0 +1,157 @@
+import git
+
+import log
+import repos
+import worktree
+import wptmeta
+
+from env import Environment
+from base import CommitBuilder
+from lock import mut, MutGuard
+
+
+env = Environment()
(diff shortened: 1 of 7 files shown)


jgraham:
I added some docstrings.

PR #566, opened 2019-11-05
Don't run aarch64 for initial landing try push.

jgraham:
We are commonly waiting 5+ hours for these results, which delays landing
by a long time. Just doing a single run in this configuration should be enough


jgraham:
I just realised this probably won't actually work because we run the metadata update with `--full`, so we would delete all the `aarch64` conditions. So we either need to drop this or change that.

jgraham:
Or maybe it's OK because we don't delete conditions that aren't reachable from the current set of runs? I need to experiment.

dhdavvie:
Closing this until we find a better solution

PR #567, opened 2019-11-08
Fix getting metadata

jgraham:
The HTTP endpoints for this changed a little, so update the code to match.

### DIFF
--- a/sync/wptfyi.py
+++ b/sync/wptfyi.py
@@ -3,6 +3,8 @@ import time
 import urllib
 import urlparse
 
+from six import iteritems
+
 
 WPT_FYI_BASE = "https://wpt.fyi/api/"
 STAGING_HOST = "staging.wpt.fyi"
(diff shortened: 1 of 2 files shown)


sync/wptfyi.py
```diff
@@ -104,22 +106,27 @@ def get_results(run_ids, test=None, query=None, staging=False):
...
         for product in products:
             url.add_query("product", product)
-    if link:
-        body = {"exists": [{"link": link}]}
-    else:
-        body = None
+    resp = requests.get(url.build())
 
-    if body is not None:
-        resp = requests.post(url.build(), json=body)
```
dhdavvie: Do we no longer have this use case?

jgraham: This never worked the way it was supposed to; the `search` api was returning the wrong things. Now we just do the same thing in client code (but this will all go away once we have the metadata repo available locally).

PR #569, opened 2019-11-08
Remove reference to expired try pushes

jgraham:


dhdavvie:
This doesn't seem like something that should be removed? We can still have expired try pushes to my knowledge

jgraham:
No, they got removed and it broke this command.

dhdavvie:
Ahh yes I remember now!

PR #571, opened 2019-11-12
Support getting data from old TaskCluster

jgraham:
When migrating to the new taskcluster, I forgot that we still needed read access to the
old taskcluster for try jobs that have already run. This patch adds a check that a given taskId
is accessible on each TaskCluster instance before deciding which queue or index to use.

### DIFF
--- a/sync/tc.py
+++ b/sync/tc.py
@@ -17,9 +17,12 @@ from errors import RetryableError
 from threadexecutor import ThreadExecutor
 
 
-TASKCLUSTER_ROOT_URL = "https://firefox-ci-tc.services.mozilla.com/"
+TASKCLUSTER_ROOT_URL = "https://firefox-ci-tc.services.mozilla.com"
 QUEUE_BASE = TASKCLUSTER_ROOT_URL + "/api/queue/v1/"
 INDEX_BASE = TASKCLUSTER_ROOT_URL + "/api/index/v1/"
+OLD_TASKCLUSTER_ROOT_URL = "https://taskcluster.net"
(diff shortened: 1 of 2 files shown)


sync/tc.py
```diff
@@ -34,12 +37,37 @@
...
+    if task_id not in tc_base_cache:
+        try:
+            resp = requests.get(QUEUE_BASE + "task/%s" % task_id)
+            resp.raise_for_status()
+        except requests.HTTPError:
+            try:
+                resp = requests.get(OLD_QUEUE_BASE + "task/%s" % task_id)
+                resp.raise_for_status()
+            except requests.HTTPError:
+                value = defaults
```
dhdavvie: Is this the right action to take here? I realise this is very edge case but if the tc base is the old one but there was a network issue, then this will get cached as being on the new taskcluster.

jgraham: OK, updated.

sync/tc.py
```diff
@@ -354,10 +388,21 @@ def is_status_fn(statuses):
...
+        resp = requests.get(idx_url)
+        try:
+            resp.raise_for_status()
+        except requests.HTTPError as e:
+            error = e
+            continue
+        else:
+            error = None
+        idx = resp.json()
+        task_id = idx.get("taskId")
```
dhdavvie: Shouldn't these two lines be in the else block? Currently it looks like if it passes on the first try, it'll then try with the second index base, notice the error and then raise it further down in the `if error:` block.

jgraham: Good catch, thanks!

PR #572, opened 2019-11-13
Unify notifications for wpt.fyi and Gecko CI data

jgraham:
Create a single object holding all of the data used for notifications; this
will be reused for deciding which bugs to file. Create a unified notification message
using the data from both sources instead of sepearating them out into two distinct
messages


jgraham:
I might need to remove some of the sample data since it's rather large.

jgraham:
I believe so.

PR #573, opened 2019-11-15
Improve the error handling for updating try pushes after landing.

jgraham:
Don't abort early in case we have a complete try push, and ensure that
in the case of error we always add a comment to the bug so that it's
clear something has gone wrong.

### DIFF
--- a/sync/command.py
+++ b/sync/command.py
@@ -262,6 +262,7 @@ def do_detail(git_gecko, git_wpt, sync_type, obj_id, *args, **kwargs):
 
 
 def do_landing(git_gecko, git_wpt, *args, **kwargs):
+    import errors
     import landing
     import update
     current_landing = landing.current(git_gecko, git_wpt)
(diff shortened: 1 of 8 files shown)


sync/command.py
```diff
@@ -303,13 +303,7 @@ def update_landing():
...
                     update_landing()
                     if current_landing.latest_try_push == try_push:
-                        landing.try_push_complete(git_gecko,
-                                                  git_wpt,
-                                                  try_push,
-                                                  current_landing,
-                                                  allow_push=kwargs["push"],
-                                                  accept_failures=accept_failures,
-                                                  tasks=tasks)
+                        landing.push_to_gecko(git_gecko, git_wpt, current_landing)
```
dhdavvie: I'm not sure why we would skip `try_push_complete` here, maybe I'm missing something?

jgraham: If the try push is marked as complete we must have already run the `try_push_complete` logic on that try push without hitting an error. So we can immediately do the `push_to_gecko` step without rerunning the whole function on the try push.

sync/landing.py
```diff
@@ -1023,17 +1007,58 @@ def try_push_complete(git_gecko, git_wpt, try_push, sync, allow_push=True,
...
 
+def needinfo_users():
+    needinfo_users = [item.strip() for item in
+                      (env.config["gecko"]["needinfo"]
+                       .get("landing", "")
+                       .split(","))]
+    return [item for item in needinfo_users if item]
+
+
+def record_failure(sync, log_msg, bug_msg, fixup_msg=None):
```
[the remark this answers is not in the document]
jgraham: Perhaps, although they're all handled errors, and the subsequent `AbortError` will show up in newrelic.

jgraham:
@dhdavvie This is updated now FYI

PR #575, opened 2019-11-18
Enable pushing to phabriactor using moz-phab

jgraham:
Now that inbound is gone, we have to push to autoland. But pushing to
autoland is hard because we struggle to find a time when it's open and
we don't hit a conflict with lando. Instead what we'd like is to end
up in the lando queue, but so far lando only knows how to read from
phabricator. So as a temporary solution we adopt the following
process:

* Push the landing to Phabricator using moz-phab
* Needinfo the wpt sync user to manually push the commits to lando and
  invoke lando (ignoring all the warnings about unlanded code).

### DIFF
--- a/config/dev/sync.ini
+++ b/config/dev/sync.ini
@@ -32,6 +32,7 @@ logs.max-count = 20
 try.max-tests = 500
 try.stability_count = 5
 upstream.needinfo-users=example@example.org,
+landing.needinfo-users=example@example.org,
 
 [web-platform-tests]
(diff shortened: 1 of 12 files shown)


jgraham:
Note that this depends on dependencies of https://bugzilla.mozilla.org/show_bug.cgi?id=1596130

dhdavvie:
Do we want to rebase this and merge it to master?

jgraham:
I think we would need a way to pref it off

jgraham:
I don't think there's a need to do this at the moment.

PR #577, opened 2019-11-21
[WIP] Phabricator Event Listening

dhdavvie:
This uses the Phabricator API to poll the stories/events, filter out the unnecessary ones, and pass the useful ones onto the Celery worker. 

Currently this listener can be run like so: `./bin/run_docker_dev.sh run --worker --phab`, but I am not sure what the best way to proceed in the future if we want the listener to be on as part of the deployment process. I've tested it locally in Docker, and can get it listening and placing messages into the RabbitMQ queue for celery. 

The token required is a Phabricator token, locally I have been running using my own. Created at https://phabricator.services.mozilla.com/settings/user/USERNAME/page/apitokens/

I also want to add some more newrelic metrics into this, so that we can see how often it triggers and generally how it runs via the dashboard.

There is also no persistence at the moment, so there is no way to restart from where the listener left off before shutdown. A work around would be to write the latest event to a file.


sync/phabricator/listen.py
```diff
@@ -0,0 +1,121 @@
...
+                 "retitled D",  # Maybe useful to upstream info?
+                 "blocking reviewer(s) for D",
+                 "planned changes to D",
+                 "resigned from D"]
+
+    event_mapping = {
+        "updated the diff for D": "commit",
+        "created D": "commit",
+        "closed D": "closed",
+        "abandoned D": "closed",  # Bit unsure about this one
```
jgraham: I think the difference is that abandoned means that this isn't going to land, whereas closed happens when it does land? So I guess we want to treat those cases separately.

sync/phabricator/listen.py
```diff
@@ -0,0 +1,121 @@
...
+            before = int(self.latest['chronologicalKey'])
+
+        done = False
+        feed = []
+
+        def chrono_key(feed_story_tuple):
+            return int(feed_story_tuple[1]["chronologicalKey"])
+
+        # keep fetching stories from Phabricator until there are no more stories to fetch
+        while not done:
```
jgraham: I think this might be clearer with a `while True:/break` setup, but I'm not sure.

sync/phabricator/listen.py
```diff
@@ -0,0 +1,121 @@
...
+                    continue
+            done = True
+        return feed
+
+    def parse(self, feed):
+        # Go through rows in reverse order, and ignore first row as it has the table headers
+        for event in feed:
+
+            # Split the text to get the part that describes the event type
+            event_text = re.compile("[0-9]{5}:").split(event['text'])[0]
```
jgraham: Is there a docs link to the event feed format we could use here?

sync/phabricator/listen.py
```diff
@@ -0,0 +1,121 @@
...
+
+            # Check if this is an event we wish to ignore
+            if any([event_type in event_text for event_type in PhabEventListener.blacklist]):
```
jgraham: You don't need the `[]`; it avoids constructing an intermediate list.

sync/phabricator/listen.py
```diff
@@ -0,0 +1,121 @@
...
+            if event['type'] is None:
+                continue
+
+            # Add the event to the queue, and set this as the latest parsed
+            handle.apply_async(("phabricator", event))
+            self.latest = event
+
+    @staticmethod
+    def map_event_type(event_text):
+        for event_type, mapping in PhabEventListener.event_mapping.items():
```
jgraham: I wonder if it's faster enough to compile these things into a regex. Not one for now but a possibility to bear in mind.

dhdavvie: Worth considering, will put a comment in the code to keep this in mind

sync/phabricator/listen.py
dhdavvie: ahh yes, nice catch

sync/phabricator/listen.py
dhdavvie: Not that I am aware of, this endpoint is still unstable and has little documentation, and I can't seem to find any documentation regarding Feed events. Will try asking the Phabricator channel

sync/phabricator/listen.py
dhdavvie: That makes sense, I agree we should probably be handling that differently

sync/phabricator/listen.py
```diff
@@ -0,0 +1,121 @@
...
+from phabricator import Phabricator
+
+from .. import log
+from ..tasks import handle
+
+
+logger = log.get_logger(__name__)
+
+
+class PhabEventListener:
```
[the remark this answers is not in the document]
dhdavvie: ahh yes I forget this is a thing in Python 2

sync/phabricator/listen.py
dhdavvie: I'll look into alternative setups

jgraham:
Are you planning to make more changes here, or should I re-review?

dhdavvie:
I have one or two more things I will address, will refresh the review
request when those are pushed :)

dhdavvie:
I think writing a test for this would be a good idea, I'll get to working on that. The area I would like feedback on is how should we approach launching this? I have added some code to start_wptsync.sh so that this listener can be started with a flag, but it feels like there should be a more complete process for it?

jgraham:
Tests look good thanks. Maybe we should check the actual list of events returned rather than just the list length.

dhdavvie:
Yes I think thats a good idea, will update that on the next push

PR #582, opened 2019-11-28
Don't mark failures as unacceptable if there are retriggers

jgraham:
Presumably in this case we already accepted the failures when the
retriggers were scheduled. Of course this could break if it happens
that someone manually retriggered some jobs, but that seems like a
smaller risk.

### DIFF
--- a/sync/landing.py
+++ b/sync/landing.py
@@ -689,7 +689,7 @@ MANUAL PUSH: wpt sync bot
 
         target_success_rate = 0.5 if not try_push.stability else 0.8
 
-        if try_push.infra_fail:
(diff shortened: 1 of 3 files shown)


dhdavvie:
Are we certain this is the case? it seems bad to assume that the errors were accepted earlier

jgraham:
I guess the other option is to store a bit of state on the try push indicating that we accepted the failures. That might be better.

dhdavvie:
Personally I think thats the better way to go, as long as it is not too much work.

jgraham:
Yes, agreed :)

jgraham:
I've repushed with an extra test, but I assume it doesn't need another round of review.

PR #586, opened 2019-12-04
File bugs for failing tests

jgraham:
This is the initial phase of this work and will probably initially only be turned on for `Testing :: web-platform-tests` components. For other components I may add some in-tree metadata that determines what to do for each component so it can be self-serve.

### DIFF
--- a/sync/notify/bugs.py
+++ b/sync/notify/bugs.py
@@ -1,0 +1,228 @@
+import json
+import re
+import os
+import subprocess
+from collections import defaultdict
+
+import newrelic
+from six import iteritems, itervalues
+
+from .msg import detail_part
+from .. import log
+from ..bugcomponents import components_for_wpt_paths
+from ..env import Environment
+from ..lock import mut
+from ..meta import Metadata
+from ..projectutil import Mach
+
+logger = log.get_logger(__name__)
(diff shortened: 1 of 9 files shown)


sync/notify/bugs.py
```diff
@@ -0,0 +1,228 @@
...
+            component = components_by_path[test_path_by_id[test_id]]
+            test_results_by_component[component].append((test_id, subtest, test_result))
+
+        for component, test_results in iteritems(test_results_by_component):
+            if component == "UNKNOWN":
+                # For things with no component don't file a bug
+                continue
+
+            if require_opt_in:
+                opt_in_components = [item.strip()
```
dhdavvie: bit nit picky, but wouldn't it be better to generate this once before the loop instead of during every iteration? This is assuming it doesn't change between iterations

sync/wptmeta/__init__.py
```diff
@@ -137,15 +153,21 @@ def __init__(self, reader, writer):
 
     def iterlinks(self, test_id, product=None, subtest=None, status=None):
         """Get the metadata for a specific test"""
```
dhdavvie: This no longer seems to be true for all cases now that test_id can be None

PR #600, opened 2019-12-13
Detect commit events in Phab listener

dhdavvie:


sync/phab/listen.py
```diff
@@ -9,6 +9,9 @@
 
 logger = log.get_logger(__name__)
 
+RE_EVENT = re.compile("[0-9]{5,}:")
+RE_COMMIT = re.compile("(committed|accepted) r[A-Z]+[a-z0-9]+:")
```
jgraham: `[a-f0-9]+` I think?

dhdavvie: Ahh yes of course, thanks!

PR #606, opened 2020-01-03
Better handle failed decision tasks

jgraham:
This prevents a number of errors we experience when the decision task fails. It doesn't quite provide the optimal solution to the problem which is to rebase and start a new try push (since usually a failed decision task is a problem in the gecko code).

### DIFF
--- a/sync/tc.py
+++ b/sync/tc.py
@@ -25,6 +25,8 @@ OLD_QUEUE_BASE = "https://queue.taskcluster.net/v1/"
 OLD_INDEX_BASE = "https://queue.taskcluster.net/v1/"
 _DATE_FMT = "%Y-%m-%dT%H:%M:%S.%fZ"
 
+TREEHERDER_BASE = "https://treeherder.mozilla.org/"
+
 SUCCESS = "completed"
 FAIL = "failed"
 EXCEPTION = "exception"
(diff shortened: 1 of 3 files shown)


sync/tc.py
```diff
@@ -411,6 +413,24 @@ def lookup_index(index_name):
     return task_id
 
 
+def lookup_treeherder(project, revision):
+    push_data = fetch_json(TREEHERDER_BASE + "api/project/%s/push/" % (project,),
+                           params={"revision": revision})
+
+    push_id = push_data.get("results", {})[0].get("id")
```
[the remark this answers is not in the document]
jgraham: Uh, that doesn't make sense, because lists don't have a `.get`. But as a result the rest of this change doesn't make sense either.

sync/tc.py
```diff
@@ -411,6 +413,25 @@ def lookup_index(index_name):
...
+        return
+
+    jobs_data = fetch_json(TREEHERDER_BASE + "api/jobs/",
+                           {"push_id": push_id})
+    property_names = jobs_data["job_property_names"]
+    idx_name = property_names.index("job_type_name")
+    idx_task = property_names.index("task_id")
+    decision_tasks = [item for item in jobs_data.get("results", [])
+                      if item[idx_name] == "Gecko Decision Task"]
+    return decision_tasks[-1][idx_task]
```
dhdavvie: Can jobs_data ever not have the `results` key? If so then this could fail since `decision_tasks` would be an empty list.

jgraham: I think "no" is the real answer here. It would be much nicer is we validated the input close to the edges and then could assume correct input in the rest of the code (something, something, proper type system).

PR #614, opened 2020-01-08
Try preventing celery setting up its own logging.

jgraham:
We aren't seeing logging to sync.ini, presumably because the celery logging setup
isn't interacting well with our setup. Try just disabling all the celery setup

### DIFF
--- a/sync/worker.py
+++ b/sync/worker.py
@@ -1,5 +1,14 @@
 import celery
 from celery.beat import crontab
+from celery.signals import setup_logging
+
+
+@setup_logging.connect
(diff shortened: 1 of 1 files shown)


jgraham:
I don't know if this is the right thing to do, but it seems like a thing to try in order to unbreak the logging.

jgraham:
Oh I pushed the merge button by mistake. But I think it's worth at least trying this.

dhdavvie:
It didn't fail the build at least, so lets keep an eye on it during deployment

PR #647, opened 2020-02-24
Set gecko_base when doing gecko_rebase

jgraham:
The gecko_rebase() function previously forgot to set the
gecko-base property of the sync, so the commit range would
end up being too large. This only mattered in practice when there
were other wpt-sync commits on autoland, but not on central.

Change things so that when we do the rebase we always set the new base.
There is still a failure mode where the rebase fails and we don't
manually redo the rebase, so the stored base would end up as
autoland rather than central, but the actual base would be central.
Even in this pathological case it's pretty unlikely that anything
actually bad will happen; the worst cas scenario is that there are
commits on central that haven't been applied to autoland that look
like part of the sync.

### DIFF
--- a/sync/downstream.py
+++ b/sync/downstream.py
@@ -62,7 +62,7 @@ class DownstreamSync(SyncProcess):
                           ("complete", "open")]  # Unfortunately, if a backout occurs
 
     @classmethod
-    @constructor(lambda args: ("downstream", str(args['pr_id'])))
+    @constructor(lambda args: ("downstream", str(args["pr_id"])))
     def new(cls, lock, git_gecko, git_wpt, wpt_base, pr_id, pr_title, pr_body):
         # TODO: add PR link to the comment
         sync = super(DownstreamSync, cls).new(lock,
@@ -548,14 +548,7 @@ class DownstreamSync(SyncProcess):
                     # TODO: Actually check if they landed?
                     reset_head = new_base
                 gecko_work.git.reset(reset_head, hard=True)
-                try:
-                    self.gecko_rebase(new_base)
-                except git.GitCommandError:
-                    try:
-                        gecko_work.git.rebase(abort=True)
-                    except git.GitCommandError:
-                        pass
-                    raise AbortError("Rebasing onto latest gecko failed")
+                self.gecko_rebase(new_base, abort_on_fail=True)
                 self.wpt_to_gecko_commits(base=new_base)
                 return True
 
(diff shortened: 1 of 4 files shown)


jgraham:
@karlcow

sync/downstream.py
```diff
@@ -548,14 +548,7 @@ def rebase_apply():
...
                     reset_head = new_base
                 gecko_work.git.reset(reset_head, hard=True)
-                try:
-                    self.gecko_rebase(new_base)
-                except git.GitCommandError:
-                    try:
-                        gecko_work.git.rebase(abort=True)
-                    except git.GitCommandError:
-                        pass
-                    raise AbortError("Rebasing onto latest gecko failed")
```
karlcow: This was moved in sync.py

sync/sync.py
```diff
@@ -671,11 +674,24 @@ def finish(self, status="complete"):
...
+        set_new_base = True
+        try:
+            git_worktree.git.rebase(new_base.sha1)
+        except git.GitCommandError as e:
+            if abort_on_fail:
+                set_new_base = False
+                try:
+                    git_worktree.git.rebase(abort=True)
+                except git.GitCommandError:
+                    pass
```
karlcow: if the abort rebase fails, it just goes to raise an error. 
The message is ignored with the pass. why the `try` on `git_worktree.git.rebase(abort=True)`  then?

sync/downstream.py
```diff
@@ -677,7 +670,7 @@ def wpt_to_gecko_commits(self, dependencies=None, base=None):
 
         reset_head = None
         if not keep_commits:
-            reset_head = base if base is not None else self.gecko_landing_branch()
+            reset_head = self.data['gecko-base']
```
karlcow: nit: single quote being used here.

test/test_downstream.py
```diff
@@ -289,6 +289,22 @@ def test_metadata_update(env, git_gecko, git_wpt, pull_request, pull_request_com
...
+    downstream.new_wpt_pr(git_gecko, git_wpt, pr)
+    sync = load.get_pr_sync(git_gecko, git_wpt, pr["number"])
+
+    assert len(sync.gecko_commits) == 1
+
+    with SyncLock.for_process(sync.process_name) as lock:
+        with sync.as_mut(lock):
+            assert sync.data['gecko-base'] == downstream.DownstreamSync.gecko_landing_branch()
+            sync.gecko_rebase(downstream.DownstreamSync.gecko_integration_branch())
+            assert sync.data['gecko-base'] == downstream.DownstreamSync.gecko_integration_branch()
```
karlcow: nit: two styles of quotes `'` and `"` are used in this function.
I don't know the convention of your project, maybe it's not important.

sync/sync.py
jgraham: The idea is that if we can't abort the rebase for whatever reason, we just want to raise the original error, since in either case we have a manual fixup to make, and it means the rest of the code just has a single error to handle. In practice the abort failing is rather unlikely (it probably means something more catastrophic is going on like the disk being full or similar).

PR #649, opened 2020-02-28
Update try strings for landing pushes

jgraham:
This does two things:
 * Passes --disable-target-task-filter from Bug 1618633, which means
 that we select all the jobs that might run on central, including
 things that are not part of the default try set, but not things
 that are only on try for experiments.

 * Updates the try selector to reflect the fact that the above change
 handles fission, the devedition jobs no longer exist, and we only
 need to run shippable builds on linux32 (since those are the only
 builds in that configuration).

### DIFF
--- a/test/test_landing.py
+++ b/test/test_landing.py
@@ -64,11 +64,10 @@ def test_land_try(env, git_gecko, git_wpt, git_wpt_upstream, pull_request, set_p
                                     "fuzzy",
                                     "--artifact",
                                     "-q",
-                                    "web-platform-tests !devedition !ccov !fis",
+                                    "web-platform-tests !ccov !shippable",
                                     "-q",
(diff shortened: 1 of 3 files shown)


test/test_landing.py
```diff
@@ -64,11 +64,10 @@ def test_land_try(env, git_gecko, git_wpt, git_wpt_upstream, pull_request, set_p
                                     "fuzzy",
                                     "--artifact",
                                     "-q",
-                                    "web-platform-tests !devedition !ccov !fis",
+                                    "web-platform-tests !ccov !shippable",
                                     "-q",
-                                    "web-platform-tests fis !devedition !ccov !asan !aarch64 "
-                                    "windows10 | linux64",
-                                    "--full")
+                                    "web-platform-tests linux32 shippable",
```
jgraham:
```suggestion
                                    "web-platform-tests linux-32 shippable",
```

PR #650, opened 2020-03-04
Fix some problems related to filing bugs

jgraham:
We started filing some bugs in layout components, and this raised a few issues around the bug metadata and text. This is a first cut at fixing those issues.

### DIFF
--- a/sync/bug.py
+++ b/sync/bug.py
@@ -130,9 +130,9 @@ class Bugzilla(object):
                               method='POST', json=body)
 
     def new(self, summary, comment, product, component, whiteboard=None, priority=None,
-            url=None):
+            url=None, bug_type="task"):
         bug = bugsy.Bug(self.bugzilla,
-                        type="task",
+                        type=bug_type,
                         summary=summary,
                         product=product,
                         component=component)
@@ -376,10 +376,18 @@ class MockBugzilla(Bugzilla):
         return MockBugContext(self, bug_id)
 
     def new(self, summary, comment, product, component, whiteboard=None, priority=None,
-            url=None):
-        self._log("Creating a bug in component %s :: %s\nSummary: %s\nComment: %s\n"
-                  "Whiteboard: %s\nPriority: %s URL: %s" %
-                  (product, component, summary, comment, whiteboard, priority, url))
+            url=None, bug_type="task"):
+        self._log("Creating a bug in component {product} :: {component}\nSummary: {summary}\n"
+                  "Comment: {comment}\nWhiteboard: {whiteboard}\nPriority: {priority}\n"
+                  "URL: {url}\nType: {bug_type}".format(
+                      product=product,
+                      component=component,
+                      summary=summary,
+                      comment=comment,
+                      whiteboard=whiteboard,
+                      priority=priority,
+                      url=url,
+                      bug_type=bug_type))
         if self.known_bugs:
             bug_id = self.known_bugs[-1] + 1
         else:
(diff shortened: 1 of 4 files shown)


sync/bug.py
```diff
@@ -130,9 +130,9 @@ def comment(self, bug_id, comment, **kwargs):
                               method='POST', json=body)
 
     def new(self, summary, comment, product, component, whiteboard=None, priority=None,
-            url=None):
+            url=None, bug_type="task"):
```
karlcow: when the list of arguments start to be long it might be time to start thinking of passing another type of object, such as a dictionary. to think about the future.

sync/bug.py
```diff
@@ -376,10 +376,10 @@ def bug_ctx(self, bug_id):
         return MockBugContext(self, bug_id)
 
     def new(self, summary, comment, product, component, whiteboard=None, priority=None,
-            url=None):
+            url=None, bug_type="task"):
         self._log("Creating a bug in component %s :: %s\nSummary: %s\nComment: %s\n"
-                  "Whiteboard: %s\nPriority: %s URL: %s" %
-                  (product, component, summary, comment, whiteboard, priority, url))
+                  "Whiteboard: %s\nPriority: %s\nURL: %s\nType: %s" %
+                  (product, component, summary, comment, whiteboard, priority, url, bug_type))
```
karlcow: Here we can see how a long list of arguments start to make it difficult to read. 
probably in the future a dictionary and the use of a format with proper variable names would make it more readable than a long series of `%s`

```python
BUG_CREATION_LOG = """Creating a bug in component {product} :: {component}
Summary: {summary}
Comment: {comment}
Whiteboard: {whiteboard}
Priority: {priority}
URL: {url}
Type: {bug_type}
"""

# with a dictionary to store the data
bug_info = {'product': 'something', 'component': 'something_else', }

## python 2.7
self._log(BUG_CREATION_LOG.format(**bug_info))

## python 3.2+
self._log(BUG_CREATION_LOG.format_map(bug_info))
```

sync/upstream.py
```diff
@@ -944,11 +946,19 @@ def commit_status_changed(git_gecko, git_wpt, sync, context, status, url, sha):
...
+                    commit_author = sync.gecko_commits[0].email
+                    if commit_author:
+                        bug.needinfo(commit_author)
+            except BugsyException:
+                msg = traceback.format_exc()
+                logger.warning("Failed to update bug:\n%s" % msg)
+                # Sometimes needinfos fail because emails addresses in bugzilla don't
+                # match the commits. That's non-fatal, but record the exception here in
+                # case something more unexpected happens
+                newrelic.agent.record_exception()
```
karlcow: would it be worth to record the difference of emails and associate them for future use? Or would that introduce issues?

sync/bug.py
jgraham: I'm generally not a fan of dictionaries as arguments just as a way to package some fields together; it has a few disadvantages compared to keyword arguments:
* Reduces documentation from the function signature since it ends up with a generic `options` element.
* Harder to spot typos, since they don't end up with `InvalidArgumentError` but something at the point of use
* Often requires more code to unpack the dictionary.

That said I agree there are some advantages. I think Py3's keyword-only arguments probably address most of the problems, so once we do the Py3 conversion we can probably improve things. And for really egregious cases you can always make a custom object type like you would in a language with a proper type system.

sync/upstream.py
jgraham: There's probably some way to record the fact that a particular email failed. Then I guess we'd need a manually updated map somewhere to record the bugzilla email address for each user with a different commit email address. It's possible but certainly a followup task.

PR #651, opened 2020-03-04
Don't fail if we didn't get any result statuses in the summary

jgraham:
It seems like there's an occasional issue where we fail to get result statuses in the summary
and we call max() on an empty sequence. It's not quite clear how this happens, but we can at least
avoid throwing in that case.

### DIFF
--- a/sync/notify/msg.py
+++ b/sync/notify/msg.py
@@ -115,7 +115,7 @@ def summary_message(results):
 
     data.append("## Status Summary\n")
 
-    max_width = max(len(status) for status in result_statuses)
+    max_width = len(max(result_statuses, key=len))
     for browser in browsers:
         if not job_names[browser]:
             continue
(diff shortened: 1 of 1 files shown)


sync/notify/msg.py
```diff
@@ -109,11 +109,14 @@ def summary_message(results):
...
     data[-1] += "\n"
 
+    result_statuses = [status for status in statuses if status in summary]
+    if not result_statuses:
+        return data
+
     data.append("## Status Summary\n")
 
-    max_width = max(len(status) for status in statuses if status in summary)
+    max_width = max(len(status) for status in result_statuses)
```
karlcow: There is another way to do that:

```suggestion
    max_width = len(max(result_statuses, key=len))
```

Note that it matters very much given the nano seconds nature. :)

```
% python -m timeit 'len(max(["a", "bcd", "ef", "ghij", "kl"], key=len))'            
500000 loops, best of 5: 654 nsec per loop
% python -m timeit 'max(len(status) for status in ["a", "bcd", "ef", "ghij", "kl"])'
500000 loops, best of 5: 878 nsec per loop
```

jgraham: Oh neat, I didn't know that worked.

PR #653, opened 2020-03-05
Create PRs when pushing to wpt-metadata repo

jgraham:
We were previously doing metadata updates by pushing directly to
master in (a fork of) the wpt-metadata repository. However because the
repo has CI checks that must pass before merging PRs to ensure that
the data is valid we actually want to create a PR first. This adds an
(optional) PR creation step when writing metadata.

### DIFF
--- a/test/test_notify_bugs.py
+++ b/test/test_notify_bugs.py
@@ -148,14 +148,20 @@ def test_update_metadata(env, git_gecko, git_wpt, pull_request, git_wpt_metadata
     with patch("sync.notify.bugs.components_for_wpt_paths",
                return_value={"Testing :: web-platform-tests": ["test/test.html"]}):
         with patch("sync.notify.bugs.Mach", return_value=mock_mach(None)):
-            with SyncLock.for_process(sync.process_name) as lock:
-                with sync.as_mut(lock):
-                    bug_data = bugs.for_sync(sync, results_obj)
-                    bugs.update_metadata(sync, bug_data)
+            with patch("sync.meta.Metadata.github") as mock_github:
+                with SyncLock.for_process(sync.process_name) as lock:
+                    with sync.as_mut(lock):
+                        bug_data = bugs.for_sync(sync, results_obj)
+                        bugs.update_metadata(sync, bug_data)
+
+                assert mock_github.create_pull.called
+                head = mock_github.method_calls[0].args[-1]
+
     bugs_filed = bug_data.keys()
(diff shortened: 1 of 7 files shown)


test/test_notify_bugs.py
```diff
@@ -143,14 +143,20 @@ def test_update_metadata(env, git_gecko, git_wpt, pull_request, git_wpt_metadata
...
+                with SyncLock.for_process(sync.process_name) as lock:
+                    with sync.as_mut(lock):
+                        bug_data = bugs.for_sync(sync, results_obj)
+                        bugs.update_metadata(sync, bug_data)
+
+                assert mock_github.create_pull.called
+                head = mock_github.method_calls[0].args[-1]
+
     bugs_filed = bug_data.keys()
     assert len(bugs_filed) == 1
```
karlcow: This is working, but it might not be necessary to extract the keys for computing a length.

```python
>>> bug_data = {'a': 1}
>>> bug_data.keys()
['a']
>>> len(bug_data.keys())
1
>>> len(bug_data)
1
```

test/test_notify_bugs.py
```diff
@@ -143,14 +143,20 @@ def test_update_metadata(env, git_gecko, git_wpt, pull_request, git_wpt_metadata
...
     bug = bugs_filed[0]
```
karlcow: Is there an example of `bug_data`?
because everything seems to say that it's a dictionary with only one key.

test/test_notify_bugs.py
jgraham: It does, but we end up using the list in the next line anyway.

test/test_notify_bugs.py
jgraham: It's a mapping of `{bug_id: test results}` where the test results are a list of results that were included in that bug. I've added a docstring to the `for_sync` function that hopefully makes this clearer.

karlcow: ok that was just to confim. 
Another way would be:

```python
assert len(bug_data) == 1
[bug, test_results] = bug_data.items()
assert link.url == "%s/show_bug.cgi?id=%s" % (env.bz.bz_url, bug)
```

just fyi

jgraham: It would actually be `[(bug, test_results)] = bug_data.items()` which I guess implicitly asserts that the length is 1, but I'm not sure it's obviously more clear :)

PR #657, opened 2020-03-11
Check for taskgroup completion when tasks finish

jgraham:
There is a failure mode we've encountered where we get some builds failing,
and the resulting taskgroup doesn't complete until all the unscheduled tasks
time out a day later. But we want to handle that case upfront, as soon as we
can know that the remaining tasks aren't going to run.

To do this, we process all tasks rather than just decision tasks. When a
non-decision task completes, we load all the tasks from the taskgroup and
check for completeness ignoring anything that's unscheduled. This has some
performance cost, so we are careful to only look at tasks we scheduled
ourselves.

### DIFF
--- a/sync/handlers.py
+++ b/sync/handlers.py
@@ -182,26 +182,37 @@ class PushHandler(Handler):
 
 
 class TaskHandler(Handler):
-    """Handler for the task associated with a decision task completing.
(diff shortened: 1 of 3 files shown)


jgraham:
This code is pretty badly tested, so there's some risk to taking this.

PR #663, opened 2020-03-16
Use push to central for base of logs, not merge-base

jgraham:
The code to get the baseline logs for central assumed that it
could get the central push using the merge-base of the commit with
central. But this doesn't work in the case where the sync was
rebased onto autoland and autoland merged into central. In this
case the merge-base will be the tip of autoland at the point where
we did the rebase, and that isn't usually a push head from central.

Instead we need to take the merge-base and ask hg what the central
push containing that commit was. Happily that's almost exactly the
same as code at [1] so the implementation can be mostly copied.

[1] https://searchfox.org/mozilla-central/source/testing/web-platform/manifestdownload.py#74

### DIFF
--- a/sync/notify/results.py
+++ b/sync/notify/results.py
@@ -271,7 +271,7 @@ class Results(object):
 def get_push_changeset(commit):
     url = ("https://hg.mozilla.org/mozilla-central/json-pushes?changeset=%s&version=2&tipsonly=1" %
            commit.canonical_rev)
-    headers = {'Accept': "application/json",
+    headers = {"Accept": "application/json",
                "User-Agent": "wpt-sync"}
     resp = requests.get(url, headers=headers)
     try:
@@ -282,8 +282,8 @@ def get_push_changeset(commit):
         return None
 
     result = resp.json()
-    pushes = result['pushes']
-    [changeset] = pushes.values()[0]['changesets']
+    pushes = result["pushes"]
+    [changeset] = pushes.values()[0]["changesets"]
     return changeset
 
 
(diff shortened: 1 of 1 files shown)


sync/notify/results.py
```diff
@@ -267,14 +268,38 @@ def is_browser_only(_test, _subtest, result):
         return self.iter_filter(is_browser_only)
 
 
+def get_push_changeset(commit):
+    url = ("https://hg.mozilla.org/mozilla-central/json-pushes?changeset=%s&version=2&tipsonly=1" %
+           commit.canonical_rev)
+    headers = {'Accept': "application/json",
+               "User-Agent": "wpt-sync"}
+    resp = requests.get(url, header=headers)
```
jgraham:
```suggestion
    resp = requests.get(url, headers=headers)
```

sync/notify/results.py
```diff
@@ -267,14 +268,38 @@ def is_browser_only(_test, _subtest, result):
         return self.iter_filter(is_browser_only)
 
 
+def get_push_changeset(commit):
+    url = ("https://hg.mozilla.org/mozilla-central/json-pushes?changeset=%s&version=2&tipsonly=1" %
+           commit.canonical_rev)
+    headers = {'Accept': "application/json",
+               "User-Agent": "wpt-sync"}
```
karlcow: nit: quote styles
```suggestion
    headers = {"Accept": "application/json",
               "User-Agent": "wpt-sync"}
```

sync/notify/results.py
```diff
@@ -267,14 +268,38 @@ def is_browser_only(_test, _subtest, result):
...
+    try:
+        resp.raise_for_status()
+    except requests.exceptions.RequestException:
+        if resp.status_code != 404:
+            newrelic.agent.record_exception()
+        return None
+
+    result = resp.json()
+    pushes = result['pushes']
+    [changeset] = pushes.values()[0]['changesets']
```
karlcow:
```suggestion
    pushes = result["pushes"]
    [changeset] = pushes.values()[0]["changesets"]
```

PR #675, opened 2020-03-26
Link to CI results when filing bugs for failures

jgraham:


jgraham:
https://travis-ci.org/github/mozilla/wpt-sync/builds/667871706

sync/notify/bugs.py
```diff
@@ -159,34 +159,160 @@ def for_sync(sync, results):
...
+        split_names.append(split_name)
+        test_dir = split_name[:-1]
+        for i, (a, b) in enumerate(zip(common_prefix, test_dir)):
+            if a != b:
+                common_prefix = common_prefix[:i]
+                break
+    return split_names, common_prefix
+
+
+def make_summary(test_results, prefix, max_length=255, max_tests=3):
```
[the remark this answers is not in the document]
karlcow: I see the string being built little by little and stopping each it reaches the fixed limit.

It's a very iterative process. :) 

I wonder if it would have been faster to compute the full string and then reduce it under certain conditions. I have no idea, Just thinking for now.
* extract_prefix
* count the tests.
* `{prefix}` longer than maxlength bailout. ok
* create the 
     1. `{prefix} in {common_path}  [{strings of all tests}]`. This one is not addressed in the tests. I'll make a comment on tests. This is the longest possible. if it fails go to 2 or 3 or 4  or 5
     2. `{prefix} in {common_path} [{substring_series} and {test_remain_number} others]` if it fails go to 3 or 4 or 5
     3. `{prefix} in {common_path} [{tests_number} tests]` if it fails go to 4 or 5
     4. `{prefix} in {common_path}` if it fails go to 5.
     5. `{prefix}`

I even wonder if a dumb "compute all strings possible" and just select the one which fits the maxlength is not faster. Just food for thoughts.

test/test_notify_bugs.py
```diff
@@ -195,3 +209,32 @@ def test_already_linked(env):
...
+
+
+def test_make_summary():
+    test_results = ["/a/b/c.html", "/a/b/d.html", "/a/b/e.html", "/a/b/f.html",
+                    "/a/b/g.html"]
+    test_results = [(item, None, None) for item in test_results]
+    prefix = "Test failures"
+
+    assert (bugs.make_summary(test_results, prefix) ==
+            "Test failures in /a/b/ [c.html, d.html, e.html, and 2 others]")
```
karlcow: here you just seem to miss a test for the case where 

```python
    test_results = ["/a/b/c.html", "/a/b/d.html", "/a/b/e.html"]
```
and the output is 

```
"Test failures in /a/b/ [c.html, d.html, e.html]"
```

sync/notify/bugs.py
```diff
@@ -159,34 +159,160 @@ def for_sync(sync, results):
...
+        return True
+
+    def has_capacity(self, chars):
+        return self.current_length + chars <= self.max_length
+
+    def get(self):
+        return "".join(self.data)
+
+
+def split_id(test_id):
```
karlcow: This might deserve tests. because I didn't really understood which cases you were trying to handle with `#` and `?`

```python
>>> split_id('a/b/c/d.html')
('a', 'b', 'c', 'd.html')
>>> split_id('a/b/c/d.html#foo')
('a', 'b', 'c', 'd.html#foo')
>>> split_id('a/b/c/#foo/d.html')
('a', 'b', 'c', '#foo/d.html')
```

Also there is the module urlparse that will do this for you too. 

```
>>> import urlparse
>>> urlparse.urlparse('a/b/c/d.html#foo')
ParseResult(scheme='', netloc='', path='a/b/c/d.html', params='', query='', fragment='foo')
>>> urlparse.urlparse('a/b/c/#foo/d.html')
ParseResult(scheme='', netloc='', path='a/b/c/', params='', query='', fragment='foo/d.html')
>>> urlparse.urlparse('a/b/c/d.html#foo').path
'a/b/c/d.html'
```

but not sure if it's what you want.

sync/notify/bugs.py
```diff
@@ -159,34 +159,160 @@ def for_sync(sync, results):
...
+            last = i
+            break
+    if last:
+        name = "/".join(parts[i:])
+        parts = parts[:i]
+        parts.append(name)
+    return tuple(parts)
+
+
+def get_common_prefix(test_ids):
```
karlcow: So here python saves you a bit of time.

[`os.path.commonprefix`](https://docs.python.org/2.7/library/os.path.html#os.path.commonprefix) available in python 3 too. Takes a list

```python
>>> test_results = ["/a/b/c.html", "/a/b/d.html", "/a/b/e.html", "/a/b/f.html", "/a/b/g.html"]
>>> import os.path
>>> os.path.commonprefix(test_results)
'/a/b/'
```

That should save a bit of lines.

karlcow: And just for the pleasure.
https://github.com/python/cpython/blob/fe75b62575bcfdf1c39be71c1e50257832a596db/Lib/genericpath.py#L68-L83

sync/notify/bugs.py
jgraham: I thought about using urlparse, but I'm not sure it works exactly like I want; it's important not to change the string at all, so what I really want is to just split on `/` in the path part and keep any hash or fragment part exactly the same as the input. So `foo/bar/baz.html?foo=/bar#baz` comes out as `["foo", "bar", "baz.html?foo=/bar#baz"]`. *Probably* `urlparse and then rejoining the last path part with any query or hash components does the right thing, but it's more clear that it's definitely doing the right thing if you explictly just take pieces of the input string and don't do a reconstruction step later.

sync/notify/bugs.py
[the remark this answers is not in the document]
jgraham: Computing the full string and reducing it seems appealing, but the logic still isn't linear, you can't just have a system where you add all the pieces and take one away at a time until you end up with something under the length limit because removing one part affects which other parts should remain e.g. if you have `{prefix} in {common path} [{test 1}, {test 2}, {and N others}]` then removing `{test 2}` means adjusting `N` and removing all the tests means trying an entirely different construct. So we end up with your approach which seems like it's going to be just as much logic as we already have but done in reverse. I'm not convinced it's worthwhile to spend time trying different approaches unless it's going to be a clear win.

sync/notify/bugs.py
jgraham: So that doesn't quite work I think. It's basically just a string prefix, so for `["a/b/cde/f.html", "a/b/cdg/h.html"]` it will return `"a/b/cd"` which doesn't seem like what we want. Obviously it's possible to fix that up but then it doesn't save so many lines. Also, in theory we always want to use `/` as the seperator whereas that would use the OS path separator. In practice this code likely only works on Linux anyway.

jgraham: Looks like it works if you pass in a list of tuples, which I suppose works for us, although it's a weird undocumented corner of the API.

PR #717, opened 2020-06-25
Run modernize over the code

jgraham:
This provides initial work on compatibility with Python 3

### DIFF
--- a/sync/base.py
+++ b/sync/base.py
@@ -5,13 +5,15 @@ from collections import defaultdict, Mapping
 
 import git
 import pygit2
+import six
+from six import iteritems, itervalues
 
 from . import log
 from . import commit as sync_commit
 from .env import Environment
 from .lock import MutGuard, RepoLock, mut, constructor
 from .repos import pygit2_get
-import six
+
 
 env = Environment()
 
@@ -182,7 +184,7 @@ class ProcessNameIndex(six.with_metaclass(IdentityMap, object)):
             if isinstance(item, set):
                 rv |= item
             else:
-                stack.extend(six.itervalues(item))
+                stack.extend(itervalues(item))
 
         return rv
 
@@ -458,7 +460,7 @@ class CommitBuilder(object):
 
     def add_tree(self, tree):
         self.has_changes = True
-        for path, data in six.iteritems(tree):
+        for path, data in iteritems(tree):
             blob = self.pygit2_repo.create_blob(data)
             index_entry = pygit2.IndexEntry(path, blob, pygit2.GIT_FILEMODE_BLOB)
             self.index.add(index_entry)
@@ -604,7 +606,7 @@ class ProcessData(six.with_metaclass(IdentityMap, object)):
         return self._data.get(key, default)
 
     def items(self):
-        for key, value in six.iteritems(self._data):
+        for key, value in iteritems(self._data):
             yield key, value
 
     @mut()
--- a/sync/base.py
+++ b/sync/base.py
@@ -1,3 +1,4 @@
+from __future__ import absolute_import
 import json
 import weakref
 from collections import defaultdict, Mapping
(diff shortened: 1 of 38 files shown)


karlcow:
@jgraham do you need to keep the python 2 compatibility?

jgraham:
Yes, it needs to keep working in Python 2; we are some way away from actually supporting 3.

sync/gh.py
```diff
@@ -97,7 +99,7 @@ def remove_labels(self, pr_id, *labels):
                     newrelic.agent.record_exception()
 
     def _convert_pr_id(self, pr_id):
-        if not isinstance(pr_id, (int, long)):
+        if not isinstance(pr_id, six.integer_types):
```
karlcow, line 102: same comment than earlier about

```
from builtins import bytes, str
```

after future import

sync/landing.py
```diff
@@ -1038,14 +1039,19 @@ def update_landing(git_gecko, git_wpt, prev_wpt_head=None, new_wpt_head=None,
...
     return landing
 
 
+def retrigger():
+    try:
+        # Avoid circular import
+        from . import tasks
+        tasks.retrigger.apply_async()
+    except OperationalError:
+        logger.warning("Failed to retrigger blocked syncs")
```
karlcow, line 1052: This seems to be completely unrelated to the conversion. Or maybe because of the import?

sync/listen.py
```diff
@@ -188,13 +190,14 @@ def accept(self, body):
...
-    event_filters = {item: lambda x: True for item in handlers.GitHubHandler.dispatch_event.keys()}
+    event_filters = {item: lambda x: True
+                     for item in handlers.GitHubHandler.dispatch_event.keys()}
     event_filters["status"] = lambda x: x["payload"]["context"] != "upstream/gecko"
     event_filters["push"] = lambda x: x["payload"]["ref"] == "refs/heads/master"
 
     def __init__(self, config, logger):
         super(GitHubFilter, self).__init__(config, logger)
-        repo_path = urlparse.urlparse(config["web-platform-tests"]["repo"]["url"]).path
+        repo_path = urllib.parse.urlparse(config["web-platform-tests"]["repo"]["url"]).path
```
karlcow, line 200: This looks a better way to do it than before. In other parts of the code, the import to `six` spray `six.` keywords into the code instead of the ready to go python 3 value.

It could be a reminder to remove them when full python 3 conversion is reached or like here just the import line to change at the top.

sync/landing.py
jgraham: Yeah, we need to avoid the circular import, and it means we need to use a different approach to mocking this in the testing, so this is required to make the tests pass again.

sync/commit.py
```diff
@@ -92,7 +94,7 @@ def __init__(self, repo, commit):
         elif hasattr(commit, "sha1"):
             # Commit subclass
             sha1 = commit.sha1
-        elif isinstance(commit, (str, unicode)):
+        elif isinstance(commit, (str, six.text_type)):
```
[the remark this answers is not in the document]
jgraham: I think this is just bad conversion; in Py2 we are accepting both here so I think it makes sense to still accept both until we figure out what the actual types are here.

sync/gh.py
jgraham: This seems OK as is to me.

sync/base.py
```diff
@@ -1,15 +1,17 @@
...
-import log
-import commit as sync_commit
-from env import Environment
-from lock import MutGuard, RepoLock, mut, constructor
-from repos import pygit2_get
+from . import log
+from . import commit as sync_commit
+from .env import Environment
+from .lock import MutGuard, RepoLock, mut, constructor
+from .repos import pygit2_get
```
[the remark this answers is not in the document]
jgraham: The `from __future__ import absolute_imports` forces you into the Python 3 import syntax, even though as you say the cleaned up imports are relative. I think this could be something to clean up in the future rather than in in this PR.

sync/commit.py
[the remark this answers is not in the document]
karlcow: yes the issue I was raising was not about having both types :) it was about the form and the logical order.

PR #718, opened 2020-07-07
Allow creating a Python 3 docker image

jgraham:


jgraham:
The scope of this increased somewhat beyond what I anticipated. I added a long commit message, repeated here for ease of access:

    This adds a --py3 flag to the `run_docker_dev.sh` script that means
    the image is built with Python 3 as the wptsync Python. This can then
    be used to test and run wptsync under Python 3. The Python 3 image is
    given a different docker tag to dintinguish it from the Py 2 image.
    
    A further flag --test was added to `run_docker_dev.sh` which skips the
    ssh key setup steps. This makes the script suitable for local testing
    or testing on travis where ssh keys aren't available (or required).
    
    Note that because the shell script is terrible, the flags must be
    passed in order.
    
    As part of the Dockerfile changes, a significant refactor was
    performed. In particular the wpt-sync code is no longer copied into
    the image, but must be mounted on the host. This was the configuration
    we were using in practice anyway, and making this change improves the
    docker build speed considerably.
    
    As a result of this, various files required to set up the rest of the
    image have to be explictly copied into the image at build
    time (notably requirements files).
    
    In addition, the docker image was changed to remove the
    virtualenv. Since we have control of the full python setup inside the
    image, also adding a virtualenv was unnecessary additional complexity.
    
    Some additional simplification of the dockerfile came along for the ride.

jgraham:
@mjzffr I know you aren't working on this anymore, but since you did most of this in the first place, I'd appreciate you glancing over the changes for evidence of insanity.

mjzffr:
@jgraham acknowledged, I'll take a look soon

bin/run_docker_dev.sh
```diff
@@ -37,31 +62,32 @@ if [[ $command == "build" ]]; then
...
-         wptsync_dev --test $@
+         --mount type=bind,source=$(pwd)/test,target=/app/wpt-sync/test \
+         ${DOCKER_TAG} --test $@
 elif [[ $command == "run" ]]; then
     exec docker run -it --add-host=rabbitmq:127.0.0.1 \
     --env WPTSYNC_CONFIG=${WPTSYNC_CONFIG:-/app/config/dev/sync.ini} \
     --env WPTSYNC_CREDS=${WPTSYNC_CREDS:-/app/config/dev/credentials.ini} \
     --env WPTSYNC_GECKO_CONFIG=${WPTSYNC_GECKO_CONFIG:-/app/config/gecko_config} \
     --env WPTSYNC_WPT_CONFIG=${WPTSYNC_WPT_CONFIG:-/app/config/wpt_config} \
+    --env WPTSYNC_WPT_METADATA_CONFIG=${WPTSYNC_WPT_CONFIG:-/app/config/wpt-metadata_config} \
```
mjzffr, line 83: Possible typo?

```suggestion
    --env WPTSYNC_WPT_METADATA_CONFIG=${WPTSYNC_WPT_METADATA_CONFIG:-/app/config/wpt-metadata_config} \
```

docker/Dockerfile.dev
```diff
@@ -75,86 +94,64 @@ RUN chmod +x /tini
...
 RUN sudo chown -R app:app /app/.ssh/
 
 # Install git-cinnabar
-WORKDIR /app/git-cinnabar
-ENV PATH=/app/git-cinnabar:$PATH
-
-RUN git clone https://github.com/glandium/git-cinnabar.git . \
-    && git checkout origin/release; \
-    pip install --user requests \
+RUN cd /app/third_party/git-cinnabar \
```
mjzffr, line 131: Distinct WORKDIR commands are preferred over `RUN cd ... && ...`

I think it's mostly a readability/style recommendation, but also each RUN command adds a new image layer so you want to use RUN for instructions that actually change the image.

jgraham: In this case we need to change the image anyway (since we actually need to install stuff) and we only care about the directory for the duration of the RUN command, so it seems less stateful to set it inside that command and not change the global state of the image. I think if we were just RUNing a single line of script here I'd be more inclined to go with seperate commands, but as it is, I don't think it actually is clearer, in spite of the best practice.

PR #719, opened 2020-07-08
Check run

jgraham:


sync/gh.py
```diff
@@ -18,6 +18,53 @@
 env = Environment()
 
 
+class CheckRun(github.GithubObject.NonCompletableGithubObject):
```
karlcow, line 21: This might be useful at least to double check
https://github.com/sigmavirus24/github3.py/issues/887

This is part of https://github3.readthedocs.io/en/master/
Which seems implemented for v.1.3.0 but the docs says  the current release is 1.2.0 

ah pypi says it's 1.3.0 https://pypi.org/project/github3.py/

Anyway just fyi

karlcow: https://github.com/sigmavirus24/github3.py/blob/9a916c44afcece5e8278e434a64904aa051e19a6/src/github3/checks.py#L375

jgraham: We're using pygithub which is a different library. The implementation here is largely copied from the internals of the status api in the library, so I'm at least reasonably sure it has the right shape.

PR #720, opened 2020-07-20
Initial work in getting mypy type annotations into the code 

jgraham:


jgraham:
Note that the point of this PR isn't to fix anything, it's just to make running mypy possible, and add some initial auto-generated type annotations. The fact that there are a few hundred errors, and lots of missing annotations isn't really surprising.

jgraham:
Thanks!

PR #728, opened 2020-09-01
Ensure we set MOZBUILD_STATE_PATH when invoking mergetool

jgraham:
This itself invokes mach so we need access to the state directory


karlcow (approved):
:+1

karlcow:
note that you may need to check the linter for python, there's a lot of cascade which might be prone to errors.

PR #731, opened 2020-09-16
Make some improvements to `wptsync landable`

jgraham:
This changes the output to be a little bit more useful in the common case. It also adds `--rebase` to `wptsync retrigger`

### DIFF
--- a/sync/command.py
+++ b/sync/command.py
@@ -195,7 +195,8 @@ def get_parser():
     parser_retrigger.add_argument("--no-downstream", action="store_false", default=True,
                                   dest="downstream", help="Don't retrigger downstream syncs")
     parser_retrigger.add_argument("--rebase", default=False, action="store_true",
-                                  help="Force downstream syncs to be rebased onto the integration branch")
+                                  help="Force downstream syncs to be rebased onto the "
+                                  "integration branch")
     parser_retrigger.set_defaults(func=do_retrigger)
 
     parser_try_push_add = subparsers.add_parser("add-try",
@@ -719,8 +720,8 @@ def do_landable(git_gecko,
                 sync = DownstreamSync.for_pr(git_gecko, git_wpt, pr)
                 assert sync is not None
                 if sync.error:
-                    msg = sync.error["message"] or ""
-                    msg = "%s (%s)" % (msg, msg.split("\n")[0])
+                    err_msg = sync.error["message"] or ""
+                    msg = "%s (%s)" % (msg, err_msg.splitlines()[0])
             print("%s: %s" % (pr, msg))
 
         print("%i PRs are unlandable:" % count)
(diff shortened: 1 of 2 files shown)


sync/command.py
```diff
@@ -689,6 +695,10 @@ def do_landable(git_gecko,
         count = 0
         for pr, _, status in unlandable:
             count += 1
+            if blocked and status in (LandableStatus.ready,
+                                      LandableStatus.upstream,
+                                      LandableStatus.skip):
```
karlcow, line 700: for myself, 

`LandableStatus.ready` (and others) returns a unicode string
https://github.com/mozilla/wpt-sync/blob/1f74819e59cedfa05f1f38a3cd91237f8becf90c/sync/sync.py#L250-L267

ok sake, because it's part of a non unique tuple, but could have a edge effect if editer later to one value.

```python
>>> statuses = ('not_valid', 'weird', 'ok')
>>> status = 'valid'
>>> if status in statuses:
...     print('oopsie')
... 
>>> statuses = ('not_valid', None, None)
>>> if status in statuses:
...     print('oopsie')
... 
>>> statuses = ('not_valid')
>>> if status in statuses:
...     print('oopsie')
... 
oopsie
```

while

```python
>>> statuses = ['not_valid']
>>> if status in statuses:
...     print('oopsie')
... 
```


So will not fail in this context, but can be dangerous.

sync/command.py
```diff
@@ -712,7 +722,7 @@ def do_landable(git_gecko,
                 if sync.error:
                     msg = sync.error["message"] or ""
                     msg = "%s (%s)" % (msg, msg.split("\n")[0])
-            print("%s: %s" % (pr, msg))
+            print("%s: %s" % (pr, msg.splitlines()[0]))
```
karlcow, line 725: nit: any reasons to not using `splitlines()` twice here 

```
>>> lines = "line 1\nline 2\nline 3\n"
>>> lines.split('\n')[0]
'line 1'
>>> lines.splitlines()[0]
'line 1'
```

karlcow (approved):
some minor comments more on syntax, consistency.

sync/command.py
jgraham: I'm very confused by this comment :) The enum statuses aren't strings, they're instances of the enum subtype:
```
In [10]: type(LandableStatus.ready)                                                                                                                                                                                
Out[10]: <enum 'LandableStatus'>
```

We allow getting a string from those:
```
In [11]: LandableStatus.ready.reason_str()                                                                                                                                                                         
Out[11]: 'Ready'
```
But that's only really for display.

sync/command.py
jgraham: I think what I"m doing here doesn't make any sense. Good catch :)

PR #732, opened 2020-09-18
Get status before trying to set it

jgraham:
GitHub doesn't automatically dedupe identical status checks, and we can end up
hitting limits if we don't get the status before setting it


karlcow (approved):
🏷️ label of quality

PR #733, opened 2020-09-22
Add some additional asserts when conmputing the commits to land

jgraham:
This should help ensure that we don't have a bug that causes us to drop commits.


karlcow (approved):
nothing weird.

PR #735, opened 2020-09-24
Only handle decision tasks with the decision task handler

jgraham:
We only want to set the taskgroup id from successful decision tasks.
Handling these in both the try-task handler and the decsion-task
handler means we end up setting the taskgroup id even when the
decision task failed, which then means we have the wrong
taskgroup id and can't download the test results

### DIFF
--- a/sync/listen.py
+++ b/sync/listen.py
@@ -244,13 +244,17 @@ class DecisionTaskFilter(Filter):
     name = "decision-task"
 
     def accept(self, body):
-        tags = body.get("task", {}).get("tags", {})
-        return (tags.get("kind") == "decision-task" and
-                tags.get("createdForUser") == "wptsync@mozilla.com")
+        return is_decision_task(body)
 
 
 class TryTaskFilter(Filter):
     name = "try-task"
 
     def accept(self, body):
-        return True
+        return not is_decision_task(body)
+
+
+def is_decision_task(body):
+    tags = body.get("task", {}).get("tags", {})
+    return (tags.get("kind") == "decision-task" and
+            tags.get("createdForUser") == "wptsync@mozilla.com")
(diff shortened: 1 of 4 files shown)


sync/listen.py
```diff
@@ -244,13 +244,17 @@ class DecisionTaskFilter(Filter):
...
 
 class TryTaskFilter(Filter):
     name = "try-task"
 
     def accept(self, body):
-        return True
+        return not is_decision_task(body)
+
+
+def is_decision_task(body):
```
karlcow, line 257: nit: missing docstrings but none of the code has docstrings. So not sure it matters.

sync/listen.py
```diff
@@ -244,13 +244,17 @@ class DecisionTaskFilter(Filter):
...
+    tags = body.get("task", {}).get("tags", {})
+    return (tags.get("kind") == "decision-task" and
+            tags.get("createdForUser") == "wptsync@mozilla.com")
```
karlcow, line 260:
```python
>>> body = {'task': {}}
>>> tags = body.get("task", {}).get("tags", {})
>>> tags
{}
>>> tags.get("kind") == "decision-task" and tags.get("createdForUser") == "wptsync@mozilla.com"
False
>>> body = {'task': {'tags': {}}}
>>> tags = body.get("task", {}).get("tags", {})
>>> tags
{}
>>> tags.get("kind") == "decision-task" and tags.get("createdForUser") == "wptsync@mozilla.com"
False
>>> body = {'task': {'tags': {'kind': 'decision-task', 'createdForUser': 'wptsync@mozilla.com'}}}
>>> tags = body.get("task", {}).get("tags", {})
>>> tags
{'kind': 'decision-task', 'createdForUser': 'wptsync@mozilla.com'}
>>> tags.get("kind") == "decision-task" and tags.get("createdForUser") == "wptsync@mozilla.com"
True
```

All Good!

karlcow (approved):
No issue.

sync/listen.py
jgraham: I agree with your assessment that this is easy to test, so I added one ;)

PR #736, opened 2020-09-24
Ensure has_metadata is called with bytes

jgraham:
has_metadata assumes that it's being called with a commit message that's
represented as bytes. UpstreamSync.from_pr also called it with a PR body
that's Text. Add the relevant type annotations and ensure we pass in bytes.


karlcow (approved):
good to go

PR #738, opened 2020-09-24
Don't error if a metadata update fails

jgraham:
This records the error in new relic for diagnosis, but otherwise
allows the sync to continue.

### DIFF
--- a/sync/meta.py
+++ b/sync/meta.py
@@ -1,4 +1,6 @@
 from __future__ import absolute_import
+import sys
+
 import git
 import newrelic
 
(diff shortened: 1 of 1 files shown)


sync/meta.py
```diff
@@ -160,30 +163,26 @@ def exit_mut(self):
...
                 else:
                     if self.create_pr:
-                        newrelic.agent.record_custom_event("metadata_update_create_pr", params={
-                            "branch": self.branch
-                        })
                         self.github.create_pull(message,
                                                 "Update from bug %s" % self.process_name.obj_id,
                                                 self.branch,
                                                 remote_ref)
+                    err = None
```
karlcow, line 174: This will be executed if no exception is being raise, which means that `err = None` already before the while loop. And the break after just end the loop. So not sure if err = None is necessary. Ah yes because if it was in the while loop for a previous failure it would have the `err` value of the previous iteration, which would impact the `if err:` in the code later. Gotcha.

karlcow (approved):
good to go

PR #739, opened 2020-09-24
Improve error reporting when the moz.build update fails

jgraham:
This happened a few times and it's hard to diagnose. Also
it possibly shouldn't be fatal.


karlcow (approved):
+1

PR #740, opened 2020-09-25
Fix the asserts when checking we have all commits in a landing

jgraham:
The previous patch failed to consider two cases:

 * Upstream syncs which typically (but not always) don't result in a
   patch to gecko.

 * Downstream syncs for changes that were also applied
   to gecko (this happens sometimes when people don't want to wait for
   the sync)

This fixes the two cases as follows:

 * For upstream syncs it counts the number of such syncs and adds that
   number to the relevent assert.

 * For downstream syncs it ensures that there's always a commit even
   if it's empty. This commit will be removed later when we rebase
   onto autoland.

### DIFF
--- a/sync/commit.py
+++ b/sync/commit.py
@@ -263,8 +263,15 @@ class Commit(object):
         return len(self.pygit2_commit.parent_ids) > 1
 
     @classmethod
-    def create(cls, repo, msg, metadata, author=None, amend=False):
-        # type: (Repo, bytes, Optional[Any], Optional[bytes], bool) -> Commit
+    def create(cls,
+               repo,  # type: Repo
+               msg,  # type: bytes
+               metadata,  # type: Optional[Dict[Text, Text]]
+               author=None,  # type: Optional[bytes]
+               amend=False,  # type: bool
+               allow_empty=False  # type: bool
+               ):
+        # type: (...) -> Commit
         msg = Commit.make_commit_msg(msg, metadata)
         commit_kwargs = {}  # type: Dict[str, Any]
         if amend:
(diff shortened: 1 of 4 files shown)


sync/commit.py
```diff
@@ -483,7 +495,11 @@ def _apply_patch(patch,  # type: bytes
                     logger.info(e)
             try:
                 logger.info("Creating commit")
-                return Commit.create(dest_repo, msg, None, amend=amend, author=author)
+                commit_kwargs = {}
+                if allow_empty:
+                    commit_kwargs["allow_empty"] = True
+                return Commit.create(dest_repo, msg, None, amend=amend, author=author,
+                                     **commit_kwargs)
```
karlcow, line 502: if I didn't miss anything, this block of code will return:
`{}` or `{"allow_empty": True}`

```
>>> def create(blah, allow_empty=False):
...     print(blah, allow_empty)
... 
>>> commit_kwargs = {}
>>> create('ok', **commit_kwargs)
ok False
>>> commit_kwargs = {"allow_empty": True}
>>> create('ok', **commit_kwargs)
ok True
```

But the dictionary has only one key:value, always.  AND `allow_empty` is not redefined in `_apply_patch` 
https://github.com/mozilla/wpt-sync/pull/740/files#diff-fe9d52c31c7eead9d4c80279ffa4af61R393-R407

so `allow_empty` will be already `False` if not changed in the call to `_apply_path` and `True` if explicitly given in the function.

So this seems it would just work the same.

```suggestion
                return Commit.create(dest_repo, msg, None, amend=amend, author=author, allow_empty)
```

karlcow (approved):
It looks logically correct.

sync/commit.py
jgraham: Yep, not sure why I didn't do that in the first place (can't use a positional argument after kwargs ofc).

PR #795, opened 2020-09-28
Remove support for running in Python 2

jgraham:
This just removes the support for building with Py2, it doesn't do anything about the (very) long tail of py2-isms in the code.


denschub:
`requirements/dev-py2.txt` is probably something that can be removed as well. :)

jgraham:
> `requirements/dev-py2.txt` is probably something that can be removed as well. :)

Good catch, thanks!

bin/run_docker_dev.sh
```diff
@@ -31,11 +22,7 @@ if [[ $command == "build" ]]; then
     else
         TEST=0
     fi
-    if $PYTHON3; then
-        ARGS="--build-arg PYTHON_BINARY=python3 --build-arg PIP_BINARY=pip3"
-    else
-        ARGS="--build-arg PYTHON_BINARY=python2 --build-arg PIP_BINARY=pip2"
-    fi
+    ARGS=""
```
karlcow, line 25: You can safely remove `ARGS` if you also remove the value in

https://github.com/mozilla/wpt-sync/pull/795/files#diff-792e0f8b4e3285a392bb896d8f512eeeR52

karlcow (approved):
ok this is the devops part of removing python 2. 
That's cool. 

Do you plan to remove the python2-ism in future PRs?

PR #799, opened 2020-10-02
Don't convert pr name to string for comparison

jgraham:
That means the comparison will always fail since both sides are now integers


denschub (approved):
WptCommit.or can either be int or None, so this looks fine.

PR #806, opened 2020-10-08
Don't encode attachment we send to bugzilla

jgraham:
AFAICT from bugsy this is wrong, even though you might think an attachment
is arbitary bytes. Hard to write a meaningful test for this, but if it
works we should stop seeing some errors in new relic


jgraham:
Oh this is nonsense. I'll redo it.

PR #810, opened 2020-10-13
Don't send cc_detail field when creating bugs.

jgraham:
This isn't actualy supported by bugzilla

### DIFF
--- a/sync/bug.py
+++ b/sync/bug.py
@@ -28,6 +28,10 @@ max_comment_length = 65535
 if "REOPENED" not in bugsy.bug.VALID_STATUS:
     bugsy.bug.VALID_STATUS.append("REOPENED")
 
+# Trying to send this field makes bug creation fail
+if "cc_detail" in bugsy.bug.ARRAY_TYPES:
+    bugsy.bug.ARRAY_TYPES.remove("cc_detail")
+
 
 def bz_url_from_api_url(api_url):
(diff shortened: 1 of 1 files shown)


sync/bug.py
```diff
@@ -28,6 +28,10 @@
 if "REOPENED" not in bugsy.bug.VALID_STATUS:
     bugsy.bug.VALID_STATUS.append("REOPENED")
 
+# Trying to send this field makes bug creation fail
+if "cc_detail" in bugsy.bug.ARRAY_TYPES:
```
karlcow, line 32: for my own understanding
https://bugzilla.readthedocs.io/en/latest/api/core/v1/bug.html?highlight=cc_detail#get-bug

> Array of objects containing detailed user information for each of the cc list members. To see the keys included in the user detail object, see below.

karlcow (approved):
👍

sync/bug.py
jgraham: Yeah, the underlying bug is that bugsy eagerly populates these fields with an empty array, so if you get the field and it wasn't in the response, you get an empty array rather than nothing. But when it comes to creating a bug that approach doesn't work because not all of the fields that are arrays in the response are valid in the request. So the real fix is to avoid sending fields the user didn't explictly set. But the wrokaround is just to not populate this field at all since we don't use it for anything.

PR #819, opened 2020-10-16
Update settings depends_on and blocks for new Bugsy API

jgraham:


denschub (approved):
Looks fine, and since you said it's still working.. r+

PR #887, opened 2020-12-17
Replace Travis with GitHub Actions.

denschub:
This PR replaces Travis with GitHub Actions, motivated by the upcoming changes to Travis. I'm not sure if Actions are enabled for this repo yet, someone with admin access to this repo can check and enable at https://github.com/mozilla/wpt-sync/settings/actions :)

This appears to be fairly reliable. Note, however, that I currently see a persistant failure inside the unit tests, see [this log](https://github.com/denschub/wpt-sync-gha/runs/1571270012?check_suite_focus=true#step:4:4281) for an example. I did run the suite multiple times and this one seems to be permafailing. I'm not sure if this is an issue with the tests, or if this is an issue with the environment. From the looks of it, I tend towards saying that the test or the code is broken, and the suite runs fine, so I opened this PR anyway.

The suite on Travis did run on Ubuntu 14.x, which isn't an option on GitHub. However, since the tests run inside Docker anyway, the host distribution is unlikely to ever matter, so I just specified `ubuntu-latest` now. That's currently 18.x, but will be 20.x [soon](https://github.com/actions/virtual-environments/issues/1816). I did a test run with 20.04 already and did not notice anything breaking. For the only step where it actually matters, the static analysis, I did hard-pin Python to 3.6 - so Ubuntu upgrades should never break anything as far as I can tell.

Total runtime is ~9 minutes, which is a bit slower than Travis, but I think that's fine.

Screenshot 2020-12-17 at 15 45 34

r? @jgraham

🎅


bin/run_docker_dev.sh
```diff
@@ -53,7 +53,7 @@ if [[ $command == "build" ]]; then
 sudo chown -R 10001:10001 config/dev workspace repos
 sudo chown -R 10002:10002 workspace/logs/rabbitmq"
 elif [[ $command == "test" ]]; then
-    exec docker run -it \
+    exec docker run \
```
jgraham, line 56: Did you verify that this doesn't affect running locally?

jgraham (commented):
Looks like we need to ask an org admin to enable actions.

bin/run_docker_dev.sh
denschub: The suite runs fine for me locally, as they do not require any input. Taking the tty away from the "run" is an issue, but the tests just dump things into STDOUT and never use STDIN.

denschub:
Converting this back to a draft, given I can't actually reproduce that one failure locally, but more importantly: because of https://bugzilla.mozilla.org/show_bug.cgi?id=1594423 ...

denschub:
Okay, I figured out what went wrong, and I "fixed" it. 

I don't exactly know why, but the order of the other browser's test results was inverted, causing `test_msg_both` to fail with

```
E           ### New Tests That Don't Pass
E         - /html/browsers/sandboxing/sandbox-new-execution-context.html: ERROR (Chrome: OK, Safari: OK)
E         ?                                                                              ------------
E         + /html/browsers/sandboxing/sandbox-new-execution-context.html: ERROR (Safari: OK, Chrome: OK)
E         ?                                                                      ++++++++++++
E         -   iframe with sandbox should load with new execution context: NOTRUN (Chrome: FAIL, Safari: FAIL)
E         ?                                                                               --------------
E         +   iframe with sandbox should load with new execution context: NOTRUN (Safari: FAIL, Chrome: FAIL)
E         ?                                                                       ++++++++++++++
```

My solution for that was to sort the dictionary containing the other browser results, which will always return in Chrome being first, which is what all tests assume. Not sure if that's a good solution, happy to hear feedback.

denschub:
I also, just now, set the `push` rule to only run on the `master` branch. Otherwise, for PRs that are based on a branch in this repo, every PR would trigger the tests twice: once for the push, once for the PR. That's not great.

sync/notify/msg.py
```diff
@@ -52,7 +52,8 @@ def status_str(result,  # type: Union[Result, SubtestResult, TestResult]
 
     if include_other_browser:
         other_browser_values = []
-        for other_browser, job_results in iteritems(result.statuses):
+        ordered_statuses = OrderedDict(sorted(result.statuses.items()))
```
jgraham, line 55: You don't actually need an `OrderedDict` here, right? You can just do `for other_browser, job_results in sorted(result.statuses.items())`

jgraham (approved):

sync/notify/msg.py
denschub: Oh, you are right. Fixed!

PR #954, opened 2021-03-02
Convert type comments to type annotations

KushalBeniwal:
I started working on issue #797.
If you have any suggestions please do comment and I'll work on them.

PR #969, opened 2021-03-17
Convert type comments to type annotations

### LINKED ISSUE #797: Convert type comments to type annotations
Now that we're using Py3 we can convert the `# type: Foo` comments to use the annotation syntax e.g. `foo: Foo`

KushalBeniwal:
Reopened PR #954 to fix #797

### DIFF
--- a/sync/base.py
+++ b/sync/base.py
@@ -75,7 +75,7 @@ class IdentityMap(type):
     determined based on data that forms part of the instance key rather
     than passed in explicitly."""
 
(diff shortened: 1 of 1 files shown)


KushalBeniwal:
> By the way, there is no need to do the type annotation changes by hand! There is a tool, https://github.com/ilevkivskyi/com2ann, that should do this automatically.

@denschub I reopened the PR. 
Can you please guide me on why the test cases are failing.

PR #1022, opened 2021-05-20
Don't check for landed landing unless we applied patches

jgraham:
Previously if we hit a problem applying the first patch for landing
we'd be left with just the base branch, and this code would
incorrectly determine that all the changes were already landed. We
can't land everything until the head commit is a landing commit, so
check for that.


denschub (approved):
This makes sense. The commit overview is showing failing tests, but unfortunately, the logs have expired, so it's probably good to re-trigger CI here to see what's wrong (or to verify that nothing is wrong)

PR #1023, opened 2021-05-21
Switch to supported repos for RabbitMQ/Erlang

jgraham:
See https://www.rabbitmq.com/install-debian.html


denschub (approved):
Looks good.

PR #1075, opened 2021-09-13
Decode the author name on downstream sync message

saschanaz:
Fixes the issue where the author name is wrapped by `b''`.


jgraham:
I landed this via https://github.com/mozilla/wpt-sync/pull/1077 with a small fix to ensure we don't throw during `decode`. Thanks for the PR!

PR #1109, opened 2021-09-30
Set rabbitmq consumer timeout to 10 hours

jgraham:
A recent rabbitmq update broke the sync because it set the task
timeout to 15 minutes, after which the node would disconnect and all
task processing would stop.

We can in some cases have quite long-running tasks. In particular the
way retrigger works currently is as a single task (this should perhaps
change), and that can take longer than an hour. So set the timeout to
10 hours to be reasonably sure that the node isn't going to get
shutdown halfway through a task.


config/rabbitmq.conf
```diff
@@ -44,3 +44,7 @@ log.console.level = info
 ## Log level to use when logging to the amq.rabbitmq.log exchange
 ##
 # log.exchange.level = info
+
+## Set consumer timeout high, since some tasks can be very slow.
+## This is 10 hours in milliseconds
+consumer_timeout = 36000000
```
denschub, line 50: Not sure if the newline at EOF is enforced by any style guide or whatever, but I figured I'd mention it.

denschub (approved):

PR #1117, opened 2021-10-04
Automated code cleanup

jgraham:
Run pyupgrade and com2ann on the code to update to Python 3 style/


jgraham:
@lutien this is the PR for the automatic Python upgrades. I don't think there's anything to salvage here apart from the tools that were used; `pyupgrade` and `comm2ann`. The logs hace expired, so I can't see what broke, but it might be worth rerunning one tool at a time and see what kinds of failures we get (if any).

jgraham:
Closed in favour of #1580

PR #1317, opened 2022-03-08
Enable running fission on try

jgraham:
Before we were still running the fission disabled configuration. Now that those have `nofis` in the job name, we were failing to run any try builds at all.


denschub (approved):
Looks reasonable to me!

PR #1550, opened 2022-09-30
Update to Ubuntu 20.04

jgraham:
This gives us a newer git version which should be compatible with m-c. It also allows us to remove Python 2.


denschub (approved):
This is more of a rubberstamp instead of a review. The changes look sensible to me, but I don't have a good way to test this, so let's... trust CI.

PR #1580, opened 2022-10-25
Run pyupgrade and com2ann on code

### LINKED ISSUE #796: Remove use of six
Now that we are Py3-only we can remove use of the six library and instead use the Python 3 native idioms everywhere.

lutien:
Closes #796 and #797.
A couple of things are done in this PR:
- Updated `iniconfig` and `smmap2` package hashes from source to built to fix the ci error in master
- Run `pyupgrade` and fix flake8 errors
- Run `com2ann`
- Fixed cyclic imports and flake8 errors
- Fixed some types (mostly by adding checks for `None` and `Optional` for variables with default value `None`)

### DIFF
--- a/sync/index.py
+++ b/sync/index.py
@@ -1,3 +1,4 @@
+from __future__ import annotations
 import abc
 import json
 from collections import defaultdict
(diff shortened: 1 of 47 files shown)


sync/index.py
```diff
@@ -28,15 +28,15 @@
...
+    name: Optional[Text] = None
+    key_fields: Tuple[Text, ...]
     unique = False
     value_cls: type = tuple
 
     # Overridden in subclasses using the constructor
     # This provides a kind of borg pattern where all instances of
     # the class have the same changes data
-    changes: Dict[Any, Any] = None
+    changes: Optional[Dict[Any, Any]] = None
```
jgraham: In this case the fact that we have to add `None` checks later on seems like a bug/limitation in mypy :/ We have code like
```py
target = self.changes
if target is None:
   return {}
# At this point mypy should know that `target` is Mapping[Any, Any]
for part in key:
    target = target[key]
    # Now it thinks that `target` is Optional[Mapping[Any, Any]] again,
    # when presumably it should be Any, although we could also work out the actual type
```
```

jgraham (approved):

PR #1616, opened 2022-11-21
Remove remaining uses of six

### LINKED ISSUE #1587: Remove remaining uses of six
We should be able to drop six entirely. After running pyupgrade, it seems like most of the remaining cases are `six.ensure_text`, where we should now be able to just check if we have text or bytes, without needing to handle the possibility of both.

lutien:
Closes #1587

### DIFF
--- a/sync/base.py
+++ b/sync/base.py
@@ -7,7 +7,6 @@ from collections.abc import Mapping
 
 import git
 import pygit2
-import six
 
 from . import log
 from . import commit as sync_commit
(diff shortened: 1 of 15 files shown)


jgraham (commented):
Looks great.

Can we also remove six from the requirements files?

lutien:
It looks like `six` is a dependency of `bugsy` and `click_repl`, so tests fail if I remove `six` from the requirements. Here I don't know enough about python dependency management, if I can do anything about it 🙈

jgraham (approved):
OK, that's a shame. No much we can do about it though. Let's merge this :)

PR #1654, opened 2022-12-19
Make sync creation more atomic

jgraham:
We sometimes run into a problem where we've created the data for a sync, but for whatever reason we don't complete creating the indexes or branch references for the sync.

To solve this, refactor sync creation so that things happen as follows:

* Create the branches. These do not themselves create a sync, and they are easy to update/remove if necessary.
* Create a single commit with the sync data and the initial indexes for the sync.

The latter change requires that we pass a CommitBuilder in to the SyncData and Index creation functions. This will add the data to the same commit, and will finally cause the commit to happen once we exit the outermost "with" block.

This also requires that we're able to construct the indexes without already having a SyncProcess object. So the make_key functions are updated to also allow passing in the key data in a form that we have before the sync itself is created.

### DIFF
--- a/sync/base.py
+++ b/sync/base.py
@@ -17,6 +17,7 @@ from .repos import pygit2_get
 from typing import (Any,
                     DefaultDict,
                     Iterator,
+                    Optional,
                     Set,
                     Tuple,
                     TYPE_CHECKING)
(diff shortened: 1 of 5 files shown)


jgraham:
I added a few extra conditions to existing tests. I realised that while this is enough to make the initial creation atomic, and so solve the issues we've had, there are some subsequent updates which don't end up in the initial commit. We could probably refactor more to fix that, but I don't think it's worth blocking this change on.

lutien (approved):
Makes sense to me! Thanks for adding the test 🙂

PR #1672, opened 2023-01-06
Downgrade billiard to 3.6.4.0

jgraham:
4.0.0 breaks Celery; the fix will be in the next Celery release


jgraham:
It seems like we still don't have a celery release that fixes https://github.com/celery/celery/pull/7446/files, so in the meantime we need to use an older version of Billiard or the sync doesn't start. Unfortunately tests don't catch this because we don't try to actually start Celery in any tests (we should perhaps try to fix that).

PR #1707, opened 2023-02-07
Fix the issue with bot reverting the whiteboard tags

### LINKED ISSUE #706: Web Platform Test Sync Bot reverts whiteboard changes on bugzilla.mozilla.org
Please see changes to the `Whiteboard` field in [1631032 - [wpt-sync] Sync PR 23067 - Rework WPT: /fetch/stale-while-revalidate/stale-css.html](https://bugzilla.mozilla.org/show_bug.cgi?id=1631032)

This has started recently (weeks).

lutien:
Closes #706 

In the scope of the bug:
- added an option to force update information about the bug from Bugzilla
- used the force option to get fresh information, when we get whiteboard tags, so when the bot tries to add bot specific tags it has updated state.

Not sure, if I can write a test here, since a lot is mocked here.

### DIFF
--- a/sync/bug.py
+++ b/sync/bug.py
@@ -111,8 +111,8 @@ class Bugzilla:
             return None
         return query["id"][0]
 
-    def _get_bug(self, bug_id: int) -> Bug | None:
-        if bug_id not in self.bug_cache:
+    def _get_bug(self, bug_id: int, force_update: bool = False) -> Bug | None:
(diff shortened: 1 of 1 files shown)


jgraham (approved):
I wonder if we should audit for more cases where we might depend on out of date bug data, but off the top of my head I can't think of anywhere else we are writing to arbitary mutable data (as opposed to appending a comment, or setting a flag or something).

PR #1774, opened 2023-03-24
Try to rebase downstream sync when error occurs

### LINKED ISSUE #1537: Rebase downstream sync when an error occurs
Sometimes we get into a state where we have a downstream sync on a m-c revision that's broken in some way (e.g. pushing to try fails). In that case it's annoying to have to manually rebase the revision to get it to work. Therefore when retrying a downstream PR that  doesn't yet have a try push (or maybe does have a try push, but where the taskgroup id is missing) we should automatically rebase that PR.

lutien:
Closes #1537

### DIFF
--- a/sync/downstream.py
+++ b/sync/downstream.py
@@ -181,14 +181,14 @@ class DownstreamSync(SyncProcess):
         if not self.requires_try:
             return DownstreamAction.ready
         if self.error:
-            return DownstreamAction.manual_fix
+            self.try_rebase()
 
         latest_try_push = self.latest_valid_try_push
         if (latest_try_push and not latest_try_push.taskgroup_id):
             if latest_try_push.status == "open":
                 return DownstreamAction.wait_try
             elif latest_try_push.infra_fail:
-                return DownstreamAction.manual_fix
+                self.try_rebase()
 
         assert self.pr is not None
         pr = env.gh_wpt.get_pull(self.pr)
@@ -244,6 +244,14 @@ class DownstreamSync(SyncProcess):
     @mut()
     def skip(self, value: bool) -> None:
         self.data["skip"] = value
+
+    def try_rebase(self) -> DownstreamAction:
+        try:
+            logger.info("Rebasing onto %s" % self.gecko_integration_branch())
+            self.gecko_rebase(self.gecko_integration_branch())
+            return DownstreamAction.ready
+        except AbortError:
+            return DownstreamAction.manual_fix
 
     @property
     def wpt(self) -> WPT:
--- a/sync/downstream.py
+++ b/sync/downstream.py
@@ -263,10 +263,17 @@ class DownstreamSync(SyncProcess):
             return DownstreamAction.manual_fix
         else:
             try:
-                logger.info("Rebasing onto %s" % self.gecko_integration_branch())
+                logger.info("Rebasing onto %s" % self.gecko_landing_branch())
                 self.tried_to_rebase = True
-                self.gecko_rebase(self.gecko_integration_branch(), abort_on_fail=True)
-                return None
+
+                commit_hash_before_rebase = self.gecko_commits.base.sha1
+                self.gecko_rebase(self.gecko_landing_branch(), abort_on_fail=True)
+                commit_hash_after_rebase = self.gecko_commits.base.sha1
+
+                if commit_hash_before_rebase == commit_hash_after_rebase:
+                    return DownstreamAction.manual_fix
+                else:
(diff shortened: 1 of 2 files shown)


test/test_downstream.py
```diff
@@ -246,6 +246,40 @@ def test_next_try_push_infra_fail(env, git_gecko, git_wpt, pull_request,
...
+
+                sync.data["affected-tests"] = {"testharness": ["example"]}
+
+                try_push = sync.next_try_push(try_cls=MockTryCls)
+                with try_push.as_mut(lock):
+                    try_push["taskgroup-id"] = None
+                    try_push.status = "complete"
+                    try_push.infra_fail = True
+
+                assert sync.next_action == downstream.DownstreamAction.ready
```
lutien, line 280: Not sure if I can actually check that rebase has happened

sync/downstream.py
```diff
@@ -181,14 +181,14 @@ def next_action(self) -> DownstreamAction:
         if not self.requires_try:
             return DownstreamAction.ready
         if self.error:
-            return DownstreamAction.manual_fix
+            self.try_rebase()
```
jgraham, line 184: Presumably if the rebase fails we want to return `DownstreamAction.manual_fix` at this point, but if it succeeds we want to continue with the remainder of the function (similarly below).

test/test_downstream.py
jgraham: So, you could make an actual commit to the integration branch in `git_gecko` as part of the test, and then check that the sync commits actually changed.

jgraham, line 280: So this seems wrong? If we rebase then we shouldn't be ready, we should be in a state where we need a try push or a stability try push (in practice the latter since we don't do non-stability try pushes for downstream syncs). So I'd expect the state here to be `downstream.DownstreamAction.try_push_stability`. But if we didn't actually do a rebase in the test, then we're still going to see the existing try push as valid, so I'm not totally sure how we end up with this result.

sync/downstream.py
```diff
@@ -181,14 +181,14 @@ def next_action(self) -> DownstreamAction:
...
 
         latest_try_push = self.latest_valid_try_push
         if (latest_try_push and not latest_try_push.taskgroup_id):
             if latest_try_push.status == "open":
                 return DownstreamAction.wait_try
             elif latest_try_push.infra_fail:
-                return DownstreamAction.manual_fix
+                self.try_rebase()
```
jgraham, line 191: So the other problem is that if we have a patch that actually breaks the infrastructure we don't want to keep trying to rebase it indefinitely. So we should do something like keep a bit of state on the sync to say "we've tried to automatically rebase this" and if it's already set we should falback to requiring manual fixup.

sync/downstream.py
```diff
@@ -245,6 +245,14 @@ def skip(self) -> bool:
     def skip(self, value: bool) -> None:
         self.data["skip"] = value
 
+    def try_rebase(self) -> DownstreamAction:
+        try:
+            logger.info("Rebasing onto %s" % self.gecko_integration_branch())
```
jgraham, line 250: I wonder if we're better off rebasing onto the integration branch (i.e. autoland) or onto central? Maybe the thing to do would be to check if central's been updated and if so rebase onto the new central, and if not keep the existing behaviour.

test/test_downstream.py
lutien: I've adjusted the mocking to get to `downstream.DownstreamAction.try_push_stability` in the test

lutien: Great! Added this

sync/downstream.py
lutien: I've added the flag to do it only once

sync/downstream.py
lutien: I think with this I'll need your help. I'm not sure, how to check that central it's updated

sync/downstream.py
```diff
@@ -245,6 +249,27 @@ def skip(self) -> bool:
...
+        self.data["tried_to_rebase"] = value
+
+    def try_rebase(self) -> DownstreamAction | None:
+        if self.tried_to_rebase:
+            return DownstreamAction.manual_fix
+        else:
+            try:
+                logger.info("Rebasing onto %s" % self.gecko_integration_branch())
+                self.tried_to_rebase = True
+                self.gecko_rebase(self.gecko_integration_branch(), abort_on_fail=True)
```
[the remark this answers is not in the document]
lutien: done ✅

PR #1847, opened 2023-05-15
Make install config job depend on update_config variable, also make update_config true by default

lutien:


ansible/roles/wptsync_host/defaults/main.yml
```diff
@@ -9,7 +9,7 @@ rebuild: true
 # credentials will be updated on server unless any of the environment variables
 # below are unset on the control machine
 update_credentials: "{{ lookup('env','WPT_CREDENTIALS') | default(false, true) }}"
-update_config: "{{ lookup('env','WPT_CONFIG') | default(false, true) }}"
+update_config: "{{ lookup('env','WPT_CONFIG') | default(true) }}"
```
jgraham, line 12: Looking at https://docs.ansible.com/ansible/latest/collections/ansible/builtin/env_lookup.html# it seems like we could rewrite these defaults as named parameters now? That might be easier to understand.

lutien: sure, updated

jgraham (approved):
Thanks!

PR #1944, opened 2023-08-04
Handle _lock on Mock objects

jgraham:
unittest.Mock got an _lock attribute to synchronise access to the Mock in multithreaded cases. Unfortunately this clashes with our use of _lock on syncs to ensure we don't allow more than one process to mutate a sync at once.

For the cases where we're mocking whole Sync objects in tests, just overwrite the _lock class attribute with another mock; this seems to be enough for the tests to pass again.

### DIFF
--- a/test/test_notify_bugs.py
+++ b/test/test_notify_bugs.py
@@ -207,7 +207,6 @@ def test_already_linked(env):
                  None,
                  None))
     sync = Mock()
-    sync = Mock()
     # The Mock class defines a _lock property which interferes with our one;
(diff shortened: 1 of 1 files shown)


test/test_notify_bugs.py
```diff
@@ -201,6 +207,10 @@ def test_already_linked(env):
                  None,
                  None))
     sync = Mock()
+    sync = Mock()
```
lutien, line 210: I think maybe here too much mocking 😄

lutien (approved):

PR #1980, opened 2023-09-04
Remove early returns for requires_try in DownstreamSync

jgraham:
The changes to make requires_try return False if there's no gecko commits were wrong in the case where there ought to be commits but we simply haven't applied them yet (e.g. beacuse of an error).

Remove the early return cases when requires_try is False in next_action and next_try_push.


test/test_downstream.py
```diff
@@ -485,3 +489,28 @@ def test_github_label_on_error(env, git_gecko, git_wpt, pull_request):
             sync.update_commits()
 
     assert env.gh_wpt.get_pull(pr["number"])["labels"] == []
+
+
+def test_github_next_action_on_error(env, git_gecko, git_wpt, pull_request, git_wpt_upstream,
+                                     upstream_gecko_commit, upstream_wpt_commit):
+    # Loacl conflict
```
lutien, line 496:
```suggestion
    # Local conflict
```

lutien (approved):

PR #2002, opened 2023-09-21
Raise an error if landing push failed

lutien:


jgraham (approved):
This seems fine, but one question: does the test actually depend on the change here? It looks like it doesn't actually simulate the push failing, only some other error when calling `try_push_complete`?

lutien:
> This seems fine, but one question: does the test actually depend on the change here? It looks like it doesn't actually simulate the push failing, only some other error when calling `try_push_complete`?

Yeah, you're right, it actually was failing at the point of comparing hashes :( So I had to mock a bit more, and then I discovered that this `raise_if_error` doesn't really work, so I did what we do in upstream code instead: https://github.com/mozilla/wpt-sync/blob/master/sync/upstream.py#L360-L366

jgraham (approved):

jgraham:
Thanks!

PR #2268, opened 2024-05-03
Copy config file during deployment

### LINKED ISSUE #2266: Make provision script update config files automatically
According to our deployment documentation, we have to manually copy config `ini` files so they would be updated on the server. We should make provision script do it automatically

lutien:
Closes #2266

### DIFF
--- a/bin/provision.sh
+++ b/bin/provision.sh
@@ -19,10 +19,14 @@ if [ "$#" -ne 2 ]; then
     exit 1
 fi
 
-if [ ! -z ${WPT_CONFIG} ]; then
-    if [ ! -f ${PWD}/config/prod/sync.ini ]; then
-        echo Please add a sync.ini file to ${PWD}/config/prod
+# Config file will be copied from `sync_prod.ini` and updated by default.
+# To disable it set WPT_CONFIG=false
+if [[ ${WPT_CONFIG} != "false" ]]; then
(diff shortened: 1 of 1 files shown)


bin/provision.sh
```diff
@@ -19,10 +19,14 @@ if [ "$#" -ne 2 ]; then
     exit 1
 fi
 
-if [ ! -z ${WPT_CONFIG} ]; then
-    if [ ! -f ${PWD}/config/prod/sync.ini ]; then
-        echo Please add a sync.ini file to ${PWD}/config/prod
+# Config file will be copied from `sync_prod.ini` and updated by default.
+# To disable it set WPT_CONFIG=false
+if [[ ${WPT_CONFIG} != "false" ]]; then
```
lutien, line 24: From what I've understood, the idea before was, that when we wanted to update config during deployment we should have manually copied the file to the `prod` folder and the run the provision script with env variable `WPT_CONFIG`. But recently I've changed it in ansible script to set this variable to `true` by default: https://github.com/mozilla/wpt-sync/pull/1847/files, meaning updating config by default. So I guess it also makes sense in provision script to copy the file by default and only if `WPT_CONFIG = false` do not do this (since it also means do not update it on the server).

jgraham (approved):
I wonder if we documented this elsewhere? If so that probably needs an update.

PR #2456, opened 2025-05-16
Use lando-cli instead of direct push

lutien:


jgraham (approved):

config/lando.toml
```diff
@@ -0,0 +1,4 @@
+[auth]
+
+api_token="%SECRET%"
```
jgraham, line 3: So we have `settings.py` to set these ini prod for the ini files, but I don't think that's going to work here, is it?

jgraham (approved):
Thanks!

jgraham (approved):

PR #2521, opened 2025-06-16
Install dependencies using uv

jgraham:


test/conftest.py
```diff
@@ -360,83 +379,85 @@ def inner(revs, bugs, message=None, bookmarks="mozilla/autoland"):
...
         return hg_commit(hg_gecko_upstream, message, bookmarks)
+
     return inner
 
 
 @pytest.fixture
 def gecko_worktree(env, git_gecko):
-    path = os.path.join(env.config["root"],
-                        env.config["paths"]["worktrees"],
-                        "gecko"
```
lutien, line 370: not sure if the comma was supposed to be here between "gecko" and "autoland"? so it wouldn't be merged into `geckoautoland`

test/conftest.py
```diff
@@ -360,83 +379,85 @@ def inner(revs, bugs, message=None, bookmarks="mozilla/autoland"):
...
     def inner(branch="test"):
-        path = os.path.join(env.config["root"],
-                            env.config["paths"]["worktrees"],
-                            "web-platform-tests"
-                            "test")
-        git_wpt.git.worktree("add",
-                             path,
-                             "origin/master")
+        path = os.path.join(
+            env.config["root"], env.config["paths"]["worktrees"], "web-platform-teststest"
```
lutien, line 397: Here we also probably miss the comma. I guess it's probably not so important since it's the test data, but I guess we would still like to have this data close to the production

lutien (approved):
Alright, that's a looot of changes 😄, but I think it generally looks fine. Only spotted a couple of things where maybe we want to fix

PR #2552, opened 2025-07-14
Set dependency ecosystem to uv

jgraham:
This should cause the uv.lock file to be updated


jgraham:
c.f. https://docs.astral.sh/uv/guides/integration/dependency-bots/#dependabot

lutien (approved):

PR #2576, opened 2025-07-18
Fix provision script

lutien:
Without these changes, the provision script can not find `ansible-playbook` command


jgraham (approved):
I'm a bit surprised if ansible doesn't already depend on ansible-core?

PR #2582, opened 2025-07-21
Don't add /app/wpt-sync as a volume in Dockerfile

jgraham:
We generally want to use this directory from the container rather than mounting the whole thing.


lutien (approved):
Do we then want to remove `uv sync` from `docker/start_wptsync.sh` already?

PR #2712, opened 2026-03-26
Reapply gecko commits if "gecko-commit-git" is missing.

lutien:
I think if we only need to add git hashes to open upstream syncs, we could just run update on them and reapply the gecko commits which should then contain the git hashes as well.

### DIFF
--- a/sync/upstream.py
+++ b/sync/upstream.py
@@ -267,24 +267,42 @@ class UpstreamSync(SyncProcess):
         if len(self.gecko_commits) == 0:
             return False
 
-        # Find the commits that were already upstreamed. Some gecko commits may not
-        # result in an upstream commit, if the patch has no effect. But if we find
-        # the last commit that was previously upstreamed then all earlier ones must
-        # also match.
-        upstreamed_commits = {item.sha1 for item in self.upstreamed_gecko_commits}
-        matching_commits = list(self.gecko_commits[:])
-        for gecko_commit in reversed(list(self.gecko_commits)):
-            if gecko_commit.sha1 in upstreamed_commits:
+        matching_commits: List[Commit] = list()
+
+        # Verify that all commits contain "gecko-commit-git" in the metadata.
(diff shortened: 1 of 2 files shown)


sync/upstream.py
```diff
@@ -267,24 +267,42 @@ def update_wpt_commits(self) -> bool:
...
-        matching_commits = list(self.gecko_commits[:])
-        for gecko_commit in reversed(list(self.gecko_commits)):
-            if gecko_commit.sha1 in upstreamed_commits:
+        matching_commits: List[Commit] = list()
+
+        # Verify that all commits contain "gecko-commit-git" in the metadata.
+        has_canonical_git_hash = True
+        for wpt_commit in self.wpt_commits:
+            if not wpt_commit.metadata.get("gecko-commit-git"):
+                has_canonical_git_hash = False
```
jgraham:
```suggestion
        has_canonical_git_hash = all(wpt_commit.metadata.get("gecko-commit-git") for wpt_commit in self.wpt_commits)
```

perhaps

sync/upstream.py
```diff
@@ -267,24 +267,42 @@ def update_wpt_commits(self) -> bool:
...
 
-        if len(matching_commits) == len(self.gecko_commits) == len(self.upstreamed_gecko_commits):
-            return False
+        if has_canonical_git_hash:
+            # Find the commits that were already upstreamed. Some gecko commits may not
+            # result in an upstream commit, if the patch has no effect. But if we find
+            # the last commit that was previously upstreamed then all earlier ones must
+            # also match.
+            upstreamed_commits = {item.sha1 for item in self.upstreamed_gecko_commits}
+            matching_commits = list(self.gecko_commits[:])
```
jgraham:
```suggestion
            matching_commits = list(self.gecko_commits)
```

Drive-by, but I don't think we need both `[:]` and `list`

jgraham (approved):

PR #2739, opened 2026-04-21
Try to create a bug for a downstream sync again on PR updates.

lutien:
We have a similar recovery mechanic on [a PR retrigger path](https://github.com/mozilla/wpt-sync/blob/master/sync/update.py#L197-L198), so it would be good to have it on the PR update path as well, since currently we sometimes have bugless downstream syncs in the landings.

### DIFF
--- a/sync/downstream.py
+++ b/sync/downstream.py
@@ -1231,19 +1231,25 @@ def update_pr(
             if sync.bug:
                 env.bz.set_status(sync.bug, "RESOLVED", "INVALID")
             sync.finish()
-        elif action == "closed":
-            # We are storing the wpt base as a reference
-            sync.data["wpt-base"] = base_sha
-            sync.next_try_push()
-        elif action == "reopened" or action == "open":
-            sync.status = "open"
-            sync.pr_status = "open"
-            sync.next_try_push()
-            assert sync.bug is not None
-            if sync.bug:
-                status = env.bz.get_status(sync.bug)
-                if status is not None and status[0] == "RESOLVED":
-                    env.bz.set_status(sync.bug, "REOPENED")
+        else:
+            # If a bug creation failed previously, try to create it again.
+            if sync.bug is None and sync.pr:
+                pr = env.gh_wpt.get_pull(sync.pr)
+                sync.create_bug(git_wpt, sync.pr, pr.title, pr.body or "")
+
+            if action == "closed":
+                # We are storing the wpt base as a reference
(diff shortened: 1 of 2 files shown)


sync/downstream.py
```diff
@@ -1231,19 +1231,25 @@ def update_pr(
...
-            sync.pr_status = "open"
-            sync.next_try_push()
-            assert sync.bug is not None
-            if sync.bug:
-                status = env.bz.get_status(sync.bug)
-                if status is not None and status[0] == "RESOLVED":
-                    env.bz.set_status(sync.bug, "REOPENED")
+        else:
+            # If a bug creation failed previously, try to create it again.
+            if sync.bug is None and sync.pr:
```
jgraham, line 1236: Is there any reason not to do this unconditionally as the first step (maybe in its own try/except block, so we run the rest of the code if it failed)?

sync/downstream.py
```diff
@@ -1231,19 +1231,25 @@ def update_pr(
...
+
+            if action == "closed":
+                # We are storing the wpt base as a reference
+                sync.data["wpt-base"] = base_sha
+                sync.next_try_push()
+            elif action == "reopened" or action == "open":
+                sync.status = "open"
+                sync.pr_status = "open"
+                sync.next_try_push()
+                assert sync.bug is not None
```
jgraham: I notice while I'm here that this assert seems surprising given the next line.

lutien: Yeah, I've removed it now. I think it makes more sense not to fail here.

sync/downstream.py
lutien: I basically followed the flow from a PR retrigger path, and I guess the idea was that it doesn't make sense to create the bug just to close it in case PR was closed, which I found reasonable here as well.
But I've added the try/except block.

sync/downstream.py
```diff
@@ -1231,19 +1231,28 @@ def update_pr(
...
+        else:
+            # If a bug creation failed previously, try to create it again.
+            if sync.bug is None and sync.pr:
+                try:
+                    pr = env.gh_wpt.get_pull(sync.pr)
+                    sync.create_bug(git_wpt, sync.pr, pr.title, pr.body or "")
+                except Exception:
+                    logger.error(f"The bug creation for the downstream sync with PR: {sync.pr} failed")
+
+
```
jgraham:
```suggestion

```

jgraham (approved):

PR #2748, opened 2026-04-28
Improve handling of backouts

jgraham:


sync/lando.py
```diff
@@ -1,28 +1,69 @@
...
+                raise
+            data = resp.json()
+            if not isinstance(data, dict):
+                raise ValueError(f"Expected map, got {data}")
+            return data
+        assert exc is not None
+        raise exc
+
+    def hg2git(self, hg_hash: str) -> Optional[str]:
+        data = self.get(f"/hg2git/firefox/{hg_hash}")
```
lutien: So apparently `urljoin` requires the path to not start with `/`. I've spotted it because we were getting 404s on prod.
I also had to update the `base_url` in [the config file](https://github.com/mozilla/wpt-sync/blob/master/sync_prod.ini#L110) to include the slash at the end.

lutien (changes requested):

sync/lando.py
jgraham:
```
In [1]: from urllib.parse import urljoin

In [2]: urljoin("https://example.org/", "/some/path")
Out[2]: 'https://example.org/some/path'

In [3]: urljoin("https://example.org", "/some/path")
Out[3]: 'https://example.org/some/path'

In [4]: urljoin("https://example.org/foo", "/some/path")
Out[4]: 'https://example.org/some/path'
```

This looks pretty much like it works as I'd expect?

jgraham: Oh, is the problem that we're putting `/api` in the base URL? In that case this is wrong (but also maybe we shouldn't do that :) )

lutien: Yeah, exactly

lutien (approved):

PR #2754, opened 2026-05-06
Temporary exclude accessibility tests in downstream syncs.

lutien:


sync/downstream.py
```diff
@@ -938,7 +938,10 @@ def affected_tests(self, revish: Any | None = None) -> Mapping[str, list[str]]:
                 for item_bytes in output.strip().split(b"\n"):
                     item = item_bytes.decode("utf8", "replace")
                     path, test_type = item.strip().split("\t")
-                    tests_by_type[test_type].append(path)
+
+                    # TODO: Temporary exclude accessibility tests until we fixed the mozharness.
+                    if not path.startswith("core-aam/"):
```
jgraham: I might suggest just at the end doing
```
if "aamtest" in test_types:
   del test_types["aamtest"]
```
because that would also handle the case where there's aam tests in some other directory like `infrastructure` (which I don't think there is at the moment, but there could be).

jgraham (changes requested):

jgraham (approved):

PR #2773, opened 2026-05-26
Escape angle brackets in PR titles and text when creating a downstream bug.

lutien:
Bugzilla API rejects requests to create a bug with a title or a comment containing the html tags.

We have an issue with creating bugs for PRs like:
https://github.com/web-platform-tests/wpt/pull/60110
https://github.com/web-platform-tests/wpt/pull/60046

### DIFF
--- a/sync/downstream.py
+++ b/sync/downstream.py
@@ -40,6 +40,18 @@ logger = log.get_logger(__name__)
 env = Environment()
 
 
+def _escape_pr_text(text: str) -> str:
+    # Escape HTML-tag-like substrings in PR text so the request body
+    # doesn't contain tokens that Bugzilla API rejects with HTTP 406. Bare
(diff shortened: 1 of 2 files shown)


sync/downstream.py
```diff
@@ -40,6 +40,18 @@
 env = Environment()
 
 
+def _escape_pr_text(text: str) -> str:
```
jgraham, line 44: Presumably using https://docs.python.org/3/library/html.html#html.escape would be fine unless bugzilla is doing something strange where it accepts e.g. `&lt;` in some cases but not others.

jgraham (changes requested):

sync/downstream.py
lutien: ah, alright, but I would still only apply it html-like tags. Updated ✅

lutien:
Fixed on the Bugzilla side.

PR #2839, opened 2026-08-19
Push to try via Lando

lutien:
~~Update the logic for pushing to try to:~~
~~- run `mach try fuzzy --no-push`;~~
~~- create a commit with the the data from the mach execution;~~
~~- push this commit together with the others to try via Lando API;~~
~~- poll the Lando job status endpoint until the commits have landed to get hg commit hash;~~
~~- clean up the try commit.~~

New workflow:
- generate a token:  `/downstream/<sync id>/<try push uuid>`
- run `mach try fuzzy --write-task-config --env WPTSYNC_TRY_PUSH_TOKEN=/downstream/<sync id>/<try push uuid>`;
- create a commit;
- push this commit together with the others to try via Lando API;
- poll the Lando job status endpoint for 60s to make sure the commits have landed;
- if it takes longer, wait for a pulse event, use the task id to get task info, and get the token from the task info to find the right sync and try push and update it's status;
- clean up the try commit;
- if we missed the event for landing, use `wptsync pr` or `wptsync retrigger` to query the lando job status endpoint.

### DIFF
--- a/sync/lando.py
+++ b/sync/lando.py
@@ -99,6 +99,8 @@ class MockLando(Lando):
         self.hg_to_git: dict[str, Optional[str]] = {}
         self.git_to_hg: dict[str, Optional[str]] = {}
         self.try_pushes: list[Mapping[str, Any]] = []
+        self.job_status: dict[int, str] = {}
+        self.default_job_status = "LANDED"
 
     def try_push(
         self,
@@ -119,10 +121,11 @@ class MockLando(Lando):
     def landing_job(self, job_id: int) -> Mapping[str, Any]:
         if not 0 < job_id <= len(self.try_pushes):
             raise ValueError(f"Lando has no job with id {job_id}")
+        status = self.job_status.get(job_id, self.default_job_status)
         return {
             "id": job_id,
-            "status": "LANDED",
-            "commit_id": "%040x" % job_id,
+            "status": status,
+            "commit_id": "%040x" % job_id if status == "LANDED" else None,
             "error": "",
             "url": urljoin(self.base_url, f"/landings/{job_id}"),
         }
--- a/sync/lando.py
+++ b/sync/lando.py
@@ -11,6 +11,7 @@ git2hg_cache: dict[str, str] = {}
 class Lando:
     def __init__(self, config: Mapping[str, Any]):
         self.base_url = config["lando"]["url"]
+        self.api_try_token = config["lando"]["try_api_token"]
 
     def get(self, path: str) -> Optional[Mapping[str, Any]]:
         exc = None
@@ -33,6 +34,45 @@ class Lando:
             return data
         assert exc is not None
         raise exc
+
+    def try_push(
+        self,
+        patches: list[str],
+        base_commit: str,
+        base_commit_vcs: str = "hg",
+        patch_format: str = "git-format-patch",
+        repo_name: str = "try",
+    ) -> int:
+        body = {
+            "base_commit": base_commit,
+            "base_commit_vcs": base_commit_vcs,
+            "patch_format": patch_format,
+            "patches": patches,
+            "repo_name": repo_name,
+        }
+        headers = {
+            "Authorization": f"Bearer {self.api_try_token}",
+            "Content-Type": "application/json",
+        }
+        responce = requests.post(
+            urljoin(self.base_url, "/api/try/patches"), headers=headers, json=body
+        )
+        try:
+            data = responce.json()
+        except ValueError:
+            data = None
+        if data is None or not isinstance(data.get("id"), int):
+            raise ValueError(f"Lando response missing id property: {data}")
+
+        return data["id"]
+
+    def landing_job(self, job_id: int) -> Mapping[str, Any]:
+        """Get the current state of a Lando landing job"""
+        data = self.get(f"/landing_jobs/{job_id}")
+        if data is None:
+            raise ValueError(f"Lando has no job with id {job_id}")
+
+        return data
 
     def hg2git(self, hg_hash: str) -> Optional[str]:
         data = self.get(f"/api/hg2git/firefox/{hg_hash}")
@@ -61,6 +101,37 @@ class MockLando(Lando):
         super().__init__(config)
         self.hg_to_git: dict[str, Optional[str]] = {}
         self.git_to_hg: dict[str, Optional[str]] = {}
+        self.try_pushes: list[Mapping[str, Any]] = []
+
+    def try_push(
+        self,
+        patches: list[str],
+        base_commit: str,
+        base_commit_vcs: str = "hg",
+        patch_format: str = "git-format-patch",
+        repo_name: str = "try",
+    ) -> int:
+        self.try_pushes.append(
+            {
+                "base_commit": base_commit,
+                "base_commit_vcs": base_commit_vcs,
+                "patch_format": patch_format,
+                "patches": patches,
+                "repo_name": repo_name,
+            }
+        )
+        return len(self.try_pushes)
+
+    def landing_job(self, job_id: int) -> Mapping[str, Any]:
+        if not 0 < job_id <= len(self.try_pushes):
+            raise ValueError(f"Lando has no job with id {job_id}")
+        return {
+            "id": job_id,
+            "status": "LANDED",
+            "commit_id": "%040x" % job_id,
+            "error": "",
+            "url": urljoin(self.base_url, f"/landings/{job_id}"),
+        }
 
     def hg2git(self, hg_hash: str) -> Optional[str]:
         return self.hg_to_git.get(hg_hash)
--- a/sync/lando.py
+++ b/sync/lando.py
@@ -13,10 +13,19 @@ class Lando:
         self.base_url = config["lando"]["url"]
         self.api_try_token = config["lando"]["try_api_token"]
 
-    def get(self, path: str) -> Optional[Mapping[str, Any]]:
+    def request(
+        self,
+        method: str,
+        path: str,
+        retry_count: int = 1,
+        headers: Optional[Mapping[str, str]] = None,
+        body: Optional[Mapping[str, Any]] = None,
+    ) -> Optional[Mapping[str, Any]]:
         exc = None
-        for _retry_count in range(5):
-            resp = requests.get(urljoin(self.base_url, path))
+        for _retry_count in range(retry_count):
+            resp = requests.request(
+                method, urljoin(self.base_url, path), headers=headers, json=body
+            )
             if resp.status_code == 404:
                 return None
             try:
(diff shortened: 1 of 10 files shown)


sync/trypush.py
```diff
@@ -37,7 +40,13 @@
...
 
 auth_tc = tc.TaskclusterClient()
-rev_re = re.compile("revision=(?P<rev>[0-9a-f]{40})")
+try_output_re = re.compile(
+    r"^Commit message:\n(?P<message>.*?)\n^Calculated try_task_config\.json:\n",
+    re.MULTILINE | re.DOTALL,
+)
+try_task_config_path = "try_task_config.json"
+lando_poll_interval = 10
+lando_poll_timeout = 30 * 60
```
jgraham, line 49: I don't love having these constants here; it's very unclear what they actually apply to. I think it would be just as easy to put them in the code.

sync/trypush.py
```diff
@@ -37,7 +40,13 @@
 env = Environment()
 
 auth_tc = tc.TaskclusterClient()
-rev_re = re.compile("revision=(?P<rev>[0-9a-f]{40})")
+try_output_re = re.compile(
+    r"^Commit message:\n(?P<message>.*?)\n^Calculated try_task_config\.json:\n",
+    re.MULTILINE | re.DOTALL,
+)
+try_task_config_path = "try_task_config.json"
```
jgraham, line 47: This is only used in one function, and we assume elsewhere that the value is always `try_task_config.json` (even if just in comments).

sync/lando.py
```diff
@@ -34,6 +35,45 @@ def get(self, path: str) -> Optional[Mapping[str, Any]]:
...
         raise exc
 
+    def try_push(
+        self,
+        patches: list[str],
+        base_commit: str,
+        base_commit_vcs: str = "hg",
+        patch_format: str = "git-format-patch",
+        repo_name: str = "try",
+    ) -> int:
```
jgraham, line 45: Which of these parameters are actually things that we might need to adjust? I'd guess not the patch format or the repo_name?

sync/lando.py
```diff
@@ -34,6 +35,45 @@ def get(self, path: str) -> Optional[Mapping[str, Any]]:
...
+            "base_commit_vcs": base_commit_vcs,
+            "patch_format": patch_format,
+            "patches": patches,
+            "repo_name": repo_name,
+        }
+        headers = {
+            "Authorization": f"Bearer {self.api_try_token}",
+            "Content-Type": "application/json",
+        }
+        responce = requests.post(
```
jgraham, line 57: Don't we want to generalize the `get` method to allow different http methods, rather than hardcoding the post logic here (without the retry handling)?

sync/trypush.py
```diff
@@ -96,34 +107,39 @@ def apply_hacks(self) -> None:
...
-        else:
-            try_rev = rev_match.group("rev")
-        if try_rev is None:
-            logger.error(msg)
-            raise AbortError(msg)
-        return try_rev
+    def read_try_rev(self, job_id: int) -> str:
+        """Wait for Lando to apply the patches we pushed and return the revision
+        it created on try"""
+        deadline = time.monotonic() + lando_poll_timeout
```
jgraham, line 119: So if I understand the current logic, this will block for up to 30 minutes waiting for a revision id to become available? I don't think we can do that; imagine if we're retriggering after some outage and we need to start many try jobs. Plus if the tree is closed and we end up with 30 minutes not being long enough it's not really clear what happens here.

It seems like we need to be able to handle the state where we have a job_id but don't have a revision yet, and need some trigger to re-poll all the jobs in that state. I guess we don't have pulse events from lando? But we could perhaps do it whenever there's a push to try, or on a timer.

sync/trypush.py
```diff
@@ -37,7 +40,13 @@
 env = Environment()
 
 auth_tc = tc.TaskclusterClient()
-rev_re = re.compile("revision=(?P<rev>[0-9a-f]{40})")
+try_output_re = re.compile(
```
jgraham, line 43: Hardcoding in the output here feels pretty fragile; this is going to break if that ever changes and worse we're going to have to support multiple versions since our worktrees can have different base commits.

I think this would really be easier if we made `mach try` write the file and read that.

sync/trypush.py
```diff
@@ -212,7 +243,78 @@ def _push(self) -> tuple[int, str]:
...
+            job_id = env.lando.try_push(patches, base_commit)
+        except Exception as e:
+            msg = f"Failed to push to try:\n{e}"
+            logger.error(msg)
+            raise RetryableError(AbortError(msg))
+        logger.info("Pushed to try as Lando job %s" % job_id)
+        return job_id
+
+    def commit_patches(self, base: str) -> list[bytes]:
+        revs = self.worktree.git.rev_list("--reverse", f"{base}..HEAD").splitlines()
```
jgraham, line 309: I think elsewhere we try to avoid using raw git commands but use either GitPython wrappers or pygit2 to access the internals directly. Do we actually need the raw commands just to get a list of commits on the branch?

jgraham (changes requested):
The main thing that worries me here is the blocking call to try to get the hg revision; I think we're going to need a better solution before we can land it.

sync/trypush.py
lutien: I've moved it in the function.

sync/lando.py
lutien: I've hardcoded the parameters that we're not changing now.

sync/trypush.py
lutien: I've updated to use a new `--write-task-config` flag.

sync/lando.py
lutien: Sure, I've updated it.

sync/trypush.py
lutien: Right, I've updated to use `worktree.iter_commits`

sync/trypush.py
lutien: Sure, I've hardcoded the values.

sync/trypush.py
lutien: Alright, as discussed, I've updated the PR to add the required information in a commit message of a try push.

lutien:
We might not need to hack commit messages after all and pass the token via the task config. I'm still waiting for the confirmation from ahal, I'll come back here with the update.

lutien:
Alright, ahal confirmed that we can just use `mach try fuzzy --env` to pass the required data to a task info.

sync/trypush.py
```diff
@@ -95,35 +115,39 @@ def apply_hacks(self) -> None:
...
-            logger.error(msg)
-            raise AbortError(msg)
-        return try_rev
+    def read_try_rev(self, job_id: int) -> str | None:
+        """Wait for Lando to apply the patches we pushed and return the revision
+        it created on try.
+
+        :return: The revision on try, or None if the job hasn't landed yet
+        """
+        deadline = time.monotonic() + 60
```
jgraham: It's unclear to me if we need this blocking path at all, and if we do why 60s is the correct timeout to hardcode. If we never get a try rev in 60s we're just blocking for no real reason, if we mostly get one in 10s if the tree is open maybe the timeout should be much shorter. In either case, if we have a timeout at all, it should be possible to set it from the caller, rather than hardcoding it.

sync/trypush.py
```diff
@@ -184,35 +252,86 @@ def _push(self) -> tuple[int, str]:
...
-                    all_paths.add(item)
-            max_tests = env.config["gecko"]["try"].get("max-tests")
-            if max_tests and len(paths) > max_tests:
-                logger.warning("Capping number of affected tests at %d" % max_tests)
-                paths = paths[:max_tests]
-            args.extend(paths)
+        # --write-task-config means mach only writes try_task_config.json to the root
+        # of the worktree; making the commit and pushing it to try is handled here
+        # instead
+        args.append("--write-task-config")
```
jgraham, line 275: What if we're on an older tree where this argument isn't supported? It's fine if we need to rebase in that case, but it would be nice if the bot handled it rather than requiring a human to do it. There are other examples in the code where we read the `--help` form `mach` to check if some argument is supported.

sync/trypush.py
```diff
@@ -315,6 +454,32 @@ def for_taskgroup(cls, git_gecko: Repo, taskgroup_id: str) -> Optional[Self]:
             return cls(git_gecko, process_name)
         return None
 
+    @classmethod
+    def for_task(cls, git_gecko: Repo, task: Mapping[str, Any]) -> Optional[Self]:
+        token = task.get("payload", {}).get("env", {}).get("WPTSYNC_TRY_PUSH_TOKEN")
```
jgraham, line 484: Not a change for this PR, but I think we should start trying to introduce pydantic or similar where we have these `Mapping[str, Any]` types representing some external data where we require access to specific fields.

sync/trypush.py
```diff
@@ -338,9 +503,55 @@ def try_rev(self, value: str) -> None:
         idx = TryCommitIndex(self.repo)
         if self.try_rev is not None:
             idx.delete(idx.make_key(self.try_rev), self.process_name)
-        self._data["try-rev"] = value
-        assert self.try_rev is not None
-        idx.insert(idx.make_key(self.try_rev), self.process_name)
+        self["try-rev"] = value
+        idx.insert(idx.make_key(value), self.process_name)
+
+    def poll_try_rev(self) -> str | None:
```
jgraham, line 535: Feels like this method should be marked as `@mut()` and have the caller acquire the lock, rather than trying to acquire the lock directly inside the method.

jgraham (changes requested):

sync/trypush.py
lutien: Yeah, it's hard to say what is the right timeout. I don't think it ever happens in the 10s, but just from personal observations, it seems like sometimes the 60s are enough. But yes, anyway, I've updated it to pass the timeout as an argument.

sync/trypush.py
lutien: Yeah, I agree that it would be nice. I've updated to rebase in case of `--write-task-config` and `--env` is missing.

sync/trypush.py
lutien: Updated

sync/trypush.py
```diff
@@ -95,35 +121,39 @@ def apply_hacks(self) -> None:
...
-        status, output = self._push()
-        return self.read_treeherder(status, output)
+    def push(self) -> tuple[int, str | None]:
+        """Push to try.
+
+        :return: Tuple of (Lando job id, revision on try). The revision is None if
+                 Lando hasn't landed the commits yet.
+        """
+        job_id = self._push()
+        return job_id, self.read_try_rev(job_id, 60)
```
jgraham: I still think unconditionally passing in a 60s delay here is a bad idea. It's probably OK if we're triggering a single try push, but in the case where we are trying to retrigger multiple syncs — a process which is already slow — we might end up with a minute (or more, given how the loop is implemented) of latency per sync. At the least we need to disable this for retriggers. But how bad would it be to set this to 0 always? We have to handle the case where try is closed anyway.

jgraham (changes requested):

jgraham (approved):

PR #2860, opened 2026-09-16
Match pulse try events with try jobs intead of decision jobs.

lutien:
The --env token is only added to jobs created by the decision task, so use a child job to match try pushes.


sync/handlers.py
```diff
@@ -265,6 +263,15 @@ def __call__(self, git_gecko: Repo, git_wpt: Repo, body: Dict[str, Any]) -> None
         taskgroup_id = body["status"]["taskGroupId"]
 
         try_push = trypush.TryPush.for_taskgroup(git_gecko, taskgroup_id)
+        task = None
+        if try_push is None:
+            # The --env token is only added to jobs created by the decision task,
+            # so use a child job to match pushes whose Lando revision is unknown.
+            task_id = body["status"]["taskId"]
+            task = tc.get_task(task_id)
```
jgraham, line 271: So this ends up adding a HTTP request per call here. I guess there aren't so many try pushes that's a huge problem, but it's at least somewhat noticeable. I guess there isn't enough information in the payload event to avoid it?

jgraham (approved):

